PR #15, opened 2017-07-20
build .so that does not contain hipBLAS

amcamd:
Build libhipblas-hcc.so from librocblas-hcc.so, not from librocblas-hcc.a and libtensile.a. With this change rocBLAS and Tensile are not included in libhipblas-hcc.so. This means that an application using hipblas will need to link to both libhipblas-hcc.so and librocblas-hcc.so.

Also add default configuration options in CMakeList.txt files for the common path.


amcamd:
Jenkins build fails, but we need to merge so .deb files can be distributed with apt-get

PR #21, opened 2017-08-15
gemm_batched function and tests

amcamd:
- add low performance implementation of hipblasGemmBatched. It calls hipblasGemm batch_count times.
- To get a performance improvement, hipblasGemmStridedBatched should be used in place of hipblasGemmBatched.

### DIFF
--- a/clients/common/arg_check.cpp
+++ b/clients/common/arg_check.cpp
@@ -1,0 +1,19 @@
+/* ************************************************************************
+ * Copyright 2016 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+
+#include <iostream>
+#include "hipblas.h"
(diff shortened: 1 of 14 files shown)


guacamoleo:
Let's hold off on merging in this pull request until we determine that the TensorFlow batched gemm is actually scattered in memory, and that it can't be refactored to directly call strided batched gemm itself. That's where we left it with Ben.

kknox:
Thus far, tensorflow appears to be OK with the batched strided GEMM API.  Keeping this open a little longer, but should be closed without merging after the next ROCm release

amcamd:
this pull request is behind the develop branch due to age,
replaced with new pull request for hipblasXgemmBatched

PR #33, opened 2017-10-12
add geam code and tests

amcamd:
add geam function and tests


amcamd:
merge from add-geam branch, not from master branch

PR #42, opened 2017-10-31
adding hipblasXgemmaBatched

amcamd:
Add hipblasXgemmBatched. This is a loop to hipblasXgemm for batch_count times. Performance optimizations can be added in the future.

### DIFF
--- a/clients/include/testing_gemm_batched.hpp
+++ b/clients/include/testing_gemm_batched.hpp
@@ -1,0 +1,354 @@
+/* ************************************************************************
+ * Copyright 2016 Advanced Micro Devices, Inc.
+ *
+ * ************************************************************************ */
+
+#include <sys/time.h>
+#include <stdlib.h>
+#include <iostream>
+#include <fstream>
+#include <vector>
+
+#include "hipblas.hpp"
+#include "utility.h"
+#include "cblas_interface.h"
+#include "norm.h"
+#include "unit.h"
+#include "arg_check.h"
+#include "flops.h"
+#include <typeinfo>
+
+using namespace std;
+
+/* ============================================================================================ */
+
+#define CLEANUP()                                        \
+do {                                                     \
+    for (int i = 0; i < batch_count; i++)                \
+    {                                                    \
+        if(dA_array[i]) hipFree(dA_array[i]);            \
+        if(dB_array[i]) hipFree(dB_array[i]);            \
+        if(dC1_array[i]) hipFree(dC1_array[i]);          \
+        if(dC2_array[i]) hipFree(dC2_array[i]);          \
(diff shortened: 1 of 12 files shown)


guacamoleo:
I think NV made a mistake by calling their first batched gemm "batchedGEMM", then wanting to add another batched gemm, the call them "batchedGEMM" and "stridedbatchedgemm".

I think it would better for us to name ours "pointerBatchedGemm" and "stridedBatchedGemm" so that we have clarity right from the start. There may be even more APIs in the future so I prefer being more specific in API names than general. Doing so still makes cuda code hipifiable.

guacamoleo:
And hipBLAS can use the exact same names as cuBLAS if that's what we want, but I'm suggesting we use better naming conventions in rocBLAS.

bensander:
Sounds good to me.

b-sumner:
What do you think of "random" and "sequential" vs. "pointer" and "strided"?

amcamd:
I think we should keep with de facto standard names. It is easier for users to call a de facto standard rather than have to call a new standard. If you change the name, the user does not know what else you have changed. If they are diligent they then need to study your new api, and the old api.

kknox:
Lets keep the hipBLAS naming convention similar to the original CUDA naming convention.  This isn't the place to deviate.  Having hipBLAS affords us flexibility in our naming convention in rocBLAS, so if we want to go with a better naming convention in rocBLAS, that would be the place.  I'm open to that.  The hipBLAS implementation itself would server as an example of a mapping of one call syntax to another.

library/src/hcc_detail/hipblas.cpp
```diff
@@ -342,10 +434,147 @@ int m, int n, int k,  const double *alpha, const double *A, int lda, long long b
...
+        {
+            return HIPBLAS_STATUS_ALLOC_FAILED;
+        }
+    }
+    else
+    {
+        h_alpha = *alpha;
+        h_beta= *beta;
+    }
+    for (int i = 0; i < batchCount; i++)
```
kknox: Making a note: in the future we should check to see if the input matrices are strided, and call the strided API instead.

clients/include/testing_gemm_batched.hpp
```diff
@@ -0,0 +1,354 @@
...
+#include "unit.h"
+#include "arg_check.h"
+#include "flops.h"
+#include <typeinfo>
+
+using namespace std;
+
+/* ============================================================================================ */
+
+#define CLEANUP()                                        \
```
kknox: We need to get in the habit of not using cleanup code like this.  I think most, if not all, of these resources could be wrapped up in a std::unique_ptr<>

PR #52, opened 2018-03-02
platform dependent typedefs to avoid conflicts wrt hgemm

Prasandhmcw:
use native cuda __half type on nvcc platform as of now.


amcamd:
Please submit this to a separate branch. There is no agreement that this should be in the develop branch.

PR #54, opened 2018-03-06
Use CMake modules rather than find package to deduce HIP configs

NEELMCW:
resolves #___

Summary of proposed changes:
-  Avoid need for hipConfig.cmake that would get generated currently only as part of ROCM installation 
-  Instead makes use of FindHIP.cmake module that gets shipped with HIP to set and deduce hip build and link configs
-  TODO: The install script still looks for hip_hcc and rocBLAS on NV stack. This needs to be conditionallychecked based on target platform


amcamd:
Please create your own branch. There is no agreement that this should be submitted to the develop branch.

NEELMCW:
@amdcamd OK Will do the needful ... Thanks

PR #56, opened 2018-04-23
added hipblas<T>nrm2()

ntrost57:
nrm2 functions were missing (required for iterative solvers)

### DIFF
--- a/clients/common/hipblas_template_specialization.cpp
+++ b/clients/common/hipblas_template_specialization.cpp
@@ -208,7 +208,7 @@
         return hipblasScasum(handle, n, x, incx, result);
     }
 */
-/*
+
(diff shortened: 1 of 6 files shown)


amcamd:
Can you add a test  to this file  https://github.com/ROCmSoftwarePlatform/hipBLAS/blob/develop/clients/gtest/blas1_gtest.cpp .  I see there is some test code, but it is never called. Can you also confirm that the test runs and passes with the command     hipblas-test --gtest_fiter=*nrm2*

ntrost57:
added nrm2 tests (all tests passed wit hcc and nvcc)

amcamd:
Thanks for adding tests. This pull request looks good to me.

PR #75, opened 2019-01-02
Correct type of hipblasHgemm's parameters.

jichangjichang:
Summary of proposed changes:
-  parameter A and B of hipblas gemm should be declared as const. This patch corrects type of hipblasHgemm's parameter A and parameter B


jichangjichang:
@amcamd @zaliu could you review and merge it.
PadddlePaddle project encountering build error due to the parameter type is incorrect.

PR #85, opened 2019-07-07
Added hip-clang and centos jenkins jobs

amdkila:
- Also ran clang-format through repo and added new .clang-format file

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -23,12 +23,14 @@ DisableFormat:   false
 Standard: Cpp11
 
 AccessModifierOffset: -4
-AlignAfterOpenBracket: true
+AlignAfterOpenBracket: Align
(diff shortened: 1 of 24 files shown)


akilaMD:
Looks like there's a bug when attempt to run the test executable on hip-clang. This should be good merge into develop.

akilaMD:
Closing this PR for now- will re-open later after issues with hip-clang docker image are fixed.

amdkila:
This passes all CI except hip-clang, which is due to an issue being worked on by the compiler team. Currently, the older Jenkinsfile fails all CI jobs in the docker stage. Plus with the email notifications added, this fixes the link for hipBLAS. I think we should merge this in ASAP.

PR #99, opened 2019-08-09
Standardizing library versions

saadrahim:
Standardizing library version to follow new standards. There should be 3 sets of numbers only.


saadrahim:
Radha requested that we fix this if there is a RC4. Radha will send an email asking us to merge if he wants it in.

PR #101, opened 2019-08-28
Guard condition for linking with hcc library

saiislam:
Added guard condition for linking with hcc library only when compiled with hcc compiler, and not with default c++ compiler.

Error: cmake was searching for non-existent "hcc shared library location (HCC_AM_LOCATION)" even with default c++ compiler.

Fix: Put a guard condition around hcc library linking step in CMakeLists.txt files of "samples" and "gtest".

Output: All test cases are now passing.

resolves # SWDEV-198086

Summary of proposed changes:
Add guard condition in following files:
1. hipblas/clients/samples/CMakeLists.txt
2. hipblas/clients/gtest/CMakeLists.txt

### DIFF
--- a/clients/gtest/CMakeLists.txt
+++ b/clients/gtest/CMakeLists.txt
@@ -90,8 +90,13 @@ if( NOT CUDA_FOUND )
   target_compile_definitions( hipblas-test PRIVATE __HIP_PLATFORM_HCC__ )
 
   get_target_property( HIP_HCC_LOCATION hip::hip_hcc IMPORTED_LOCATION_RELEASE )
-  get_target_property( HCC_AM_LOCATION hcc::hc_am IMPORTED_LOCATION_RELEASE )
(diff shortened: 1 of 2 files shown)


clients/gtest/CMakeLists.txt
```diff
@@ -90,8 +90,13 @@ if( NOT CUDA_FOUND )
   target_compile_definitions( hipblas-test PRIVATE __HIP_PLATFORM_HCC__ )
 
   get_target_property( HIP_HCC_LOCATION hip::hip_hcc IMPORTED_LOCATION_RELEASE )
-  get_target_property( HCC_AM_LOCATION hcc::hc_am IMPORTED_LOCATION_RELEASE )
-  target_link_libraries( hipblas-test PRIVATE ${HIP_HCC_LOCATION} ${HCC_AM_LOCATION} )
+
+  if( CMAKE_CXX_COMPILER MATCHES ".*/hcc$" )
```
aaronenyeshi: This might not always work. For HIP built on HCC, it will use hipcc, but still require the hc_am package.

PR #122, opened 2019-11-11
Batched & strided_batched additions

daineAMD:
Sorry for the big PR.

First step to match rocblas 2.10 functionality. Adds:
- scal_batched and scal_strided_batched
- copy_batched and copy_strided_batched
- dot(c)_batched and dot(c)_strided_batched
- swap_batched and swap_strided_batched
- asum_batched and asum_strided_batched
- nrm2_batched and nrm2_strided_batched
Along with corresponding unit tests.

All cuda implementations of L1 batched have been changed to return NOT_SUPPORTED rather than demoting them to non-batched versions.

Most of the changes are testing code changes except for library/src/*/hipblas.cpp and library/include/hipblas.h

A future PR (tomorrow) will include:
- ger_batched and ger_strided_batched
- syr_batched and syr_strided_batched
- gemm_batched and gemm_strided_batched
- complex gemm
- hdot and haxpy, maybe bfdot.

This would match rocblas 2.10 I believe.

### DIFF
--- a/clients/common/hipblas_template_specialization.cpp
+++ b/clients/common/hipblas_template_specialization.cpp
@@ -219,6 +219,64 @@ hipblasStatus_t hipblasSwap<hipDoubleComplex>(
     return hipblasZswap(handle, n, x, incx, y, incy);
 }
 
+// swap_batched
+template <>
+hipblasStatus_t
+    hipblasSwapBatched<float>(hipblasHandle_t handle, int n, float* x[], int incx, float* y[], int incy, int batch_count)
+{
+    return hipblasSswapBatched(handle, n, x, incx, y, incy, batch_count);
+}
+
+template <>
+hipblasStatus_t
+    hipblasSwapBatched<double>(hipblasHandle_t handle, int n, double* x[], int incx, double* y[], int incy, int batch_count)
+{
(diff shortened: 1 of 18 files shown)


daineAMD:
I see CI is failing since it's using rocBLAS 2.9.I'm guessing there hasn't been any sort of standard of using up-to-date rocBLAS for hipBLAS in the past, but since Braga asked for 2.10 matching I'm not sure if we can safely ignore CI (it passed on _my_ machine) or figure something else out.

amcamd:
In file install.sh we add rocblas to library_dependencies_ubuntu and then install rocblas with install_apt_packages "${library_dependencies_ubuntu[@]}". The install of rocblas is fast, but it gets the latest released version of rocBLAS, and this lags behind the rocBLAS in the develop branch. 

An alternative is to build and install the develop branch of rocBLAS. This would be slow. To do this you could add something like the following to install.sh

pushd .
  git clone -b develop https://github.com/ROCmSoftwarePlatform/rocBLAS.git
  cd rocBLAS
  ./install.sh -id 
popd

I think it is fine to merge your PR if it passes in your tests. If you have time you can try to modify install.sh so that it builds the develop branch of rocBLAS.

daineAMD:
Passed locally, merging. Will fix rocBLAS build in future PR.

PR #149, opened 2020-01-21
Add hip-clang fixes from master branch to develop branch

amdkila:
Similarly to rocBLAS, I merged in hipBLAS changes directly to the master branch so that hip-clang would build successfully. While we probably could wait until the next branching date without getting any merge conflicts, it's probably best to merge these changes into develop right away.


amdkila:
Actually, I can see the version is decremented. I'll cherrypick these changed in another PR.

PR #150, opened 2020-01-21
Merging hip-clang fixes (#147)

amdkila:
This applies the hot fix that was added from the master branch to the develop branch


install.sh
```diff
@@ -104,25 +105,25 @@ install_packages( )
   fi
 
   # dependencies needed for library and clients to build
-  local library_dependencies_ubuntu=( "make" "cmake-curses-gui" "pkg-config" "rocm-dev" )
```
aaronenyeshi: How come we remove dependency of rocm-dev for hip-clang? It should be part of that package.

amdkila: I've noticed that when installing rocm-dev in hip-clang containers, hcc is installed, and the old hipcc wrapper with hip-hcc runtime replaces the hip-vdi runtime. I was able to build hipBLAS without installing rocm-dev.

PR #167, opened 2020-03-05
Fix hip target to be compatible with hip-clang

amdkila:


clients/gtest/CMakeLists.txt
```diff
@@ -97,8 +97,8 @@ target_link_libraries( hipblas-test PRIVATE roc::hipblas cblas lapack ${GTEST_LI
 if( NOT CUDA_FOUND )
   target_compile_definitions( hipblas-test PRIVATE __HIP_PLATFORM_HCC__ )
 
-  get_target_property( HIP_HCC_LOCATION hip::hip_hcc IMPORTED_LOCATION_RELEASE )
-  target_link_libraries( hipblas-test PRIVATE ${HIP_HCC_LOCATION} )
+  get_target_property( HIP_LOCATION hip::device IMPORTED_LOCATION_RELEASE )
+  target_link_libraries( hipblas-test PRIVATE ${HIP_LOCATION} )
```
aaronenyeshi: I think IMPORTED_LOCATION_RELEASE is not available on hip::device. It should be best to use:
 target_link_libraries( hipblas-test PRIVATE hip::device )

amdkila: ah right, this caused problems on your rocBLAS PR. Thanks for the reminder, will fix this ASAP

saadrahim:
@pfultz2 This PR seems to be handling things for the new compiler in a manner that I don't think is robust. Do you have any ideas? The new compiler does not have the hip::hip_hcc targets.

PR #175, opened 2020-03-25
rocSOLVER support

tfalders:
Adds support for some rocSOLVER functions in order to match BLAS-like extension functions from cuBLAS. Note that only the batched versions are supported by cuBLAS.

- getrf (regular, batched, strided_batched)
- getrs (regular, batched, strided_batched)
- geqrf (regular, batched, strided_batched - floats and doubles only)

Note: Copying ipiv into tau in the batched geqrf functions could use improvement.


tfalders:
I've added the option to build without dependency on rocSOLVER. This can be done using the cmake BUILD_WITH_SOLVER=OFF option, or with the install.sh -n | --no-solver flags.

PR #181, opened 2020-04-10
Merge master into develop for ROCm 3.4

daineAMD:


install.sh
```diff
@@ -119,18 +121,42 @@ install_packages( )
...
+      library_dependencies_sles+=( "${custom_rocblas}" )
+    fi
+  fi
 
-    if [[ "${build_solver}" == true ]]; then
+  if [[ "${build_solver}" == true ]]; then
       library_dependencies_ubuntu+=( "rocsolver" )
       library_dependencies_centos+=( "rocsolver" )
       library_dependencies_fedora+=( "rocsolver" )
       library_dependencies_sles+=( "rocsolver" )
```
TorreZuk, line 159: Now I am not sure why these other dependencies aren't set for hip-clang, but if that is true solver used to be also excluded for hip-clang (was inside the if ).   Sorry but this may be question for CI who are on holiday.

daineAMD: That's true, looks like this section was inside excluded from hip-clang. Hm, I'm not sure of the implications of this either.
I just built using hip-clang on my local machine and it seemed to be ok.

TorreZuk: We can ask Akila next week

daineAMD: Sounds good. Will merge into develop for now.

PR #191, opened 2020-04-24
Make the return type explicit.

mhbliao:
- Remove unnecessary `auto` return type as only C++14 starts to deduce
  that automatically. As that header would be included in other
  projects, that unnecessary dependency should be removed or
  internalized.

resolves #___

Summary of proposed changes:
-
-
-


mhbliao:
SWDEV-231429 and several issues are blocked by this change.

PR #192, opened 2020-04-28
Make the return type explicit. (#191)

eidenyoshida:


leekillough:
This is not a full solution.

PR #196, opened 2020-05-04
privatize c++14 use

TorreZuk:
* test CI issues


leekillough:
I added your `#ifdef` in PR#193; the local C++14 options were actually unnecessary since the top-level `CMakeLists.txt` sets `CMAKE_CXX_STANDARD` to 14, and I verified that the build works (modulo the `hip_runtime_api.h` problems).

PR #197, opened 2020-05-04
Enable devtoolset7 in CentOS CI

eidenyoshida:
Dockerfile is not enabling properly, enabling right before build command.


TorreZuk:
I agree, the duct tape is getting pretty thick

PR #198, opened 2020-05-04
Make hipblas.h work with C instead of C++, and use std::complex for test arithmetic

leekillough:
Cherry-pick from PR#193.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,0 +1,30 @@
+;; This causes emacs to abide by formatting rules: No tab characters, except
+;; in Makefiles; BSD-style indendation, with a 4-character indentation offset
+;; (8 in assembly code and Makefiles); no whitespace at the end of lines; and
+;; a newline at the end of file.
+
+(
+ (nil . (
+  (require-final-newline . t)
+  (indent-tabs-mode . nil)
+  (eval add-hook 'before-save-hook 'delete-trailing-whitespace)
+  (eval add-hook 'before-save-hook (lambda () (untabify (point-min) (point-max))))
+ ) )
+ (prog-mode . (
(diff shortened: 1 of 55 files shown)


leekillough:
It's defined in the hipBLAS `CMakeLists.txt:`

https://github.com/ROCmSoftwarePlatform/hipBLAS/blob/ad196b561aaba0fcc541cf131244b98a175a0a7e/clients/gtest/CMakeLists.txt#L141

https://github.com/ROCmSoftwarePlatform/hipBLAS/blob/ad196b561aaba0fcc541cf131244b98a175a0a7e/clients/gtest/CMakeLists.txt#L172

However, it is only defined in `gtest/CMakeLists.txt.` It should probably be moved to hipBLAS main directory, to apply to both client and library builds. This applies to `develop` as well.

leekillough:
> 
> 
> It's defined in the hipBLAS `CMakeLists.txt:`
> 
> https://github.com/ROCmSoftwarePlatform/hipBLAS/blob/ad196b561aaba0fcc541cf131244b98a175a0a7e/clients/gtest/CMakeLists.txt#L141
> 
> https://github.com/ROCmSoftwarePlatform/hipBLAS/blob/ad196b561aaba0fcc541cf131244b98a175a0a7e/clients/gtest/CMakeLists.txt#L172
> 
> However, it is only defined in `gtest/CMakeLists.txt.` It should probably be moved to hipBLAS main directory, to apply to both client and library builds. This applies to `develop` as well.

Correction -- it's also defined for `library`, but not for `clients/samples`.  [PR#199](https://github.com/ROCmSoftwarePlatform/hipBLAS/pull/199) makes it get defined hipBLAS-wide or not.

PR #207, opened 2020-05-10
Fix hip-clang cmake symbols

amdkila:
Add hip::host and do not link against hip_hcc if cmake_cxx_compiler matches hip_hcc


library/src/CMakeLists.txt
```diff
@@ -70,7 +70,7 @@ if( NOT CUDA_FOUND )
 
   # Remove following when hcc is fixed; hcc emits following spurious warning ROCm v1.6.3
```
saadrahim, line 71: This old note could be removed

clients/samples/CMakeLists.txt
```diff
@@ -46,10 +46,10 @@ foreach( exe example-sscal;example-sgemm;example-sgemm-strided-batched;example-c
   target_compile_options( ${exe} PRIVATE -mf16c)
 
   if( CUSTOM_TARGET )
-    target_link_libraries( ${exe} PRIVATE hip::${CUSTOM_TARGET} )
+    target_link_libraries( ${exe} PRIVATE hip::${CUSTOM_TARGET} hip::host )
   else( )
     if( LIBAMDHIP64_LIBRARY )
-      target_link_libraries( ${exe} PRIVATE hip::amdhip64 )
+      target_link_libraries( ${exe} PRIVATE hip::amdhip64 hip::host )
```
saadrahim, line 52: do we need amdhip64?

CMakeLists.txt
```diff
@@ -82,13 +82,6 @@ find_library(
   HINTS "${CMAKE_PREFIX_PATH}/lib" )
 message("amdhip64: ${LIBAMDHIP64_LIBRARY}")
```
saadrahim, line 83: Do we need this?

amdkila: yes, we do need it to cover the case where libamdhip64.so doesn't exist on the user's stack, or if they're using a different runtime environment

PR #208, opened 2020-05-10
Make constexpr variable specializations static, to satisfy ODR

leekillough:
The C++ standard is ambiguous; there are many conflicting specs:
1. A `constexpr` or `const` _non-template_ variable has _internal linkage_, unless declared `extern`.
2. A `constexpr` or `const` _template_ variable has _external linkage_, unless declared `static`.
3. In C++17, a `constexpr` variable is implicitly `inline`; in C++14, variables cannot be `inline`. `inline` variables can be defined in multiple compilation units without violating ODR. Some compilers treat C++14 `constexpr` variables as `inline` even though C++14 does not recognize `inline` variables.
4. A template specialization is not allowed to have a _storage class specifier_ like `static`, but it is unclear whether this was intended to apply to variable templates, or whether _storage class specifier_ and _linkage_ are distinct concepts as far as template specializations go.

See also https://github.com/ROCmSoftwarePlatform/rocBLAS/pull/551 and the links it refers to.

This change makes the variable template specializations static, to satisfy ODR (prevent multiple definitions), but it is unclear whether this should really be required.

### DIFF
--- a/clients/include/utility.h
+++ b/clients/include/utility.h
@@ -142,10 +142,10 @@ template <typename T>
 static constexpr bool is_complex = false;
 
 template <>
-constexpr bool is_complex<hipblasComplex> = true;
+HIPBLAS_CLANG_STATIC constexpr bool is_complex<hipblasComplex> = true;
 
 template <>
-constexpr bool is_complex<hipblasDoubleComplex> = true;
+HIPBLAS_CLANG_STATIC constexpr bool is_complex<hipblasDoubleComplex> = true;
 
 // Get base types from complex types.
 template <typename T, typename = void>
--- a/library/include/hipblas.h
+++ b/library/include/hipblas.h
@@ -18,6 +18,12 @@
 #include <hip/hip_runtime_api.h>
 #include <stdint.h>
 
+#if __clang__
+#define HIPBLAS_CLANG_STATIC static
+#else
+#define HIPBLAS_CLANG_STATIC
+#endif
+
 typedef void* hipblasHandle_t;
 
 typedef uint16_t hipblasHalf;


amdkila:
Hmmm, it looks like you may need to add a clang version check, and make it static if __clang__, else keep the constexpr variable unchanged. 
When building hipBLAS with the g++ compiler, this is what I see:

[  3%] Building CXX object clients/gtest/CMakeFiles/hipblas-test.dir/set_get_vector_gtest.cpp.o
In file included from /workspace/hipBLAS/clients/gtest/../include/hipblas_vector.hpp:9:0,
                 from /workspace/hipBLAS/clients/gtest/../include/unit.h:11,
                 from /workspace/hipBLAS/clients/gtest/../include/testing_set_get_vector.hpp:14,
                 from /workspace/hipBLAS/clients/gtest/set_get_vector_gtest.cpp:6:
/workspace/hipBLAS/clients/gtest/../include/utility.h:145:1: error: explicit template specialization cannot have a storage class
 static constexpr bool is_complex<hipblasComplex> = true;
 ^
/workspace/hipBLAS/clients/gtest/../include/utility.h:148:1: error: explicit template specialization cannot have a storage class
 static constexpr bool is_complex<hipblasDoubleComplex> = true;
 ^

amdkila:
I have a PR [here](https://github.com/ROCmSoftwarePlatform/hipBLAS/pull/210) that adds this check only if the compiler is clang. It just adds one commit to your current PR, so you can merge it into your PR and then this PR will be updated

leekillough:
> Hmmm, it looks like you may need to add a clang version check, and make it static if **clang**, else keep the constexpr variable unchanged.
> When building hipBLAS with the g++ compiler, this is what I see:
> 
> [ 3%] Building CXX object clients/gtest/CMakeFiles/hipblas-test.dir/set_get_vector_gtest.cpp.o
> In file included from /workspace/hipBLAS/clients/gtest/../include/hipblas_vector.hpp:9:0,
> from /workspace/hipBLAS/clients/gtest/../include/unit.h:11,
> from /workspace/hipBLAS/clients/gtest/../include/testing_set_get_vector.hpp:14,
> from /workspace/hipBLAS/clients/gtest/set_get_vector_gtest.cpp:6:
> /workspace/hipBLAS/clients/gtest/../include/utility.h:145:1: error: explicit template specialization cannot have a storage class
> static constexpr bool is_complex = true;
> ^
> /workspace/hipBLAS/clients/gtest/../include/utility.h:148:1: error: explicit template specialization cannot have a storage class
> static constexpr bool is_complex = true;
> ^

Yeah, that's #4 in my list of ambiguities above.
I saw that error before.

leekillough:
It appears that this is a clang bug, and our original code, which specified `static` only in the primary variable template declaration, was correct, and that `static` should not appear in specializations:

https://bugs.llvm.org/show_bug.cgi?id=35863

PR #214, opened 2020-05-11
Matching rocBLAS 3.4/3.5 functionality

daineAMD:
Should match rocBLAS 3.4/3.5 functionality. rocBLAS 3.5 isn't released yet so will remain a PR for now.

Adds:
- tbsv (all variants)
- hemm (all variants)
- symm (all variants)
- geam (complex and batched/strided_batched)

Currently doesn't test conjugate geam as this is broken in rocBLAS at the moment.


daineAMD:
CI is passing except for finding a rocBLAS Ubuntu package on CI. Merging since everything else passed.

PR #220, opened 2020-05-26
BLAS 1 Fortran for hipBLAS

daineAMD:
Adds Fortran support for BLAS 1 functions. Layout is very similar to that of rocBLAS.

- Support for BLAS 1 in Fortran
- Testing for BLAS 1. Currently everything that hipBLAS previously tested is now also tested in Fortran
- Adds a hipblasSscal sample
- Some code from BLAS2/BLAS3 made the PR but is commented out - these will be added in separate PRs (soon) to try to limit the size of these


daineAMD:
The sample doesn't currently work on CUDA - I will look into it on Monday and try to fix, or remove it for now. Hopefully will have 1/2 new PRs on Monday for the other Fortran functions as well.

PR #221, opened 2020-05-27
Check for error code in install script

amdkila:


install.sh
```diff
@@ -46,15 +46,25 @@ supported_distro( )
   esac
 }
 
+# This function is helpful for dockerfiles that do not have sudo installed, but the default user is root
```
saadrahim, line 49: Maybe correct or delete the terrible comment

PR #222, opened 2020-05-28
rocSOLVER support 2

tfalders:
Added complex support for geqrf functions and removed obsolete use of USE_DEVICE_MODE macro. geqrf_batched can now call geqrf_ptr_batched in rocSOLVER, which matches the cuBLAS method signature.

### DIFF
--- a/clients/gtest/geqrf_batched_gtest.cpp
+++ b/clients/gtest/geqrf_batched_gtest.cpp
@@ -60,7 +60,7 @@ TEST_P(geqrf_batched_gtest, geqrf_batched_gtest_float)
 
     Arguments arg = setup_geqrf_batched_arguments(GetParam());
 
-    hipblasStatus_t status = testing_geqrf_batched<float>(arg);
+    hipblasStatus_t status = testing_geqrf_batched<float, float>(arg);
 
     if(status != HIPBLAS_STATUS_SUCCESS)
     {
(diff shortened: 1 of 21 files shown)


jzuniga-amd:
> Looks like CI is failing:
> ../../library/src/libhipblas.so.0.1: undefined reference to rocsolver_cgeqrf_ptr_batched' ../../library/src/libhipblas.so.0.1: undefined reference to rocsolver_dgeqrf_ptr_batched'
> ../../library/src/libhipblas.so.0.1: undefined reference to rocsolver_zgeqrf_ptr_batched' ../../library/src/libhipblas.so.0.1: undefined reference to rocsolver_sgeqrf_ptr_batched'
> Is that from an outdated version of rocSolver in CI?
> Also can you add tests for the complex versions added?

Yes... I don't think rocsolver is being built without hipclang in CI anymore. The built that this job is pulling is old and certainly does not contain geqrf_ptr_batched…

As for the complex tests, I agree, a few complex tests should be added.

tfalders:
I added complex tests, but there are some issues I need to address before I can finalize the PR. In debug mode, zaxpy is making a mess of the test results. I need to figure out what's going on there.

tfalders:
The test failures appear to be limited to my machine, as the tests are passing CI. Once the failure on rocm-docker is addressed it should be ready to merge, though I may also add getri to the PR once it's implemented in rocSOLVER.

clients/gtest/geqrf_gtest.cpp
```diff
@@ -82,7 +82,51 @@ TEST_P(geqrf_gtest, geqrf_gtest_double)
...
+
+    if(status != HIPBLAS_STATUS_SUCCESS)
+    {
+        if(arg.M < 0 || arg.N < 0 || arg.lda < arg.M)
+        {
+            EXPECT_EQ(HIPBLAS_STATUS_INVALID_VALUE, status);
+        }
+        else
+        {
+            EXPECT_EQ(HIPBLAS_STATUS_NOT_SUPPORTED, status); // for cuda
```
daineAMD, line 95: I think this should be `EXPECT_EQ(HIPBLAS_STATUS_SUCCESS, status);` since this is supported in CUDA - basically it should auto fail if it isn't a success.

clients/gtest/geqrf_strided_batched_gtest.cpp
```diff
@@ -82,7 +82,51 @@ TEST_P(geqrf_strided_batched_gtest, geqrf_strided_batched_gtest_double)
...
+
+    if(status != HIPBLAS_STATUS_SUCCESS)
+    {
+        if(arg.M < 0 || arg.N < 0 || arg.lda < arg.M || arg.batch_count < 0)
+        {
+            EXPECT_EQ(HIPBLAS_STATUS_INVALID_VALUE, status);
+        }
+        else
+        {
+            EXPECT_EQ(HIPBLAS_STATUS_NOT_SUPPORTED, status); // for cuda
```
daineAMD, line 95: CUDA doesn't support strided_batched correct? If so this is good :)

clients/gtest/geqrf_batched_gtest.cpp
```diff
@@ -82,7 +82,51 @@ TEST_P(geqrf_batched_gtest, geqrf_batched_gtest_double)
...
+
+    if(status != HIPBLAS_STATUS_SUCCESS)
+    {
+        if(arg.M < 0 || arg.N < 0 || arg.lda < arg.M || arg.batch_count < 0)
+        {
+            EXPECT_EQ(HIPBLAS_STATUS_INVALID_VALUE, status);
+        }
+        else
+        {
+            EXPECT_EQ(HIPBLAS_STATUS_SUCCESS, status);
```
daineAMD, line 95: It looks like CUDA supports the batched version, so this should be good.

clients/include/testing_getrs_batched.hpp
```diff
@@ -51,12 +51,12 @@ hipblasStatus_t testing_getrs_batched(Arguments argus)
     host_vector<int> hIpiv1(Ipiv_size);
     int              info;
 
-    device_batch_vector<T> bA(batch_count, A_size);
-    device_batch_vector<T> bB(batch_count, B_size);
+    device_batch_vector<T, 1> bA(batch_count, A_size);
```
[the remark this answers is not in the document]
tfalders: Originally I thought that my test failures could be coming from out of memory errors. I've removed it now.

clients/gtest/geqrf_gtest.cpp
tfalders: My understanding is that the non-batched geqrf is not supported in cuBLAS. It only has the batched version.

clients/gtest/geqrf_strided_batched_gtest.cpp
tfalders: As far as I know, yes.

PR #223, opened 2020-06-02
Adding Fortran support for BLAS 2

daineAMD:
Adds support for L2 Fortran. Also adds tests for L2 Fortran. Another PR coming soon today for L3. Everything passed locally but will wait for CI as well.


daineAMD:
Failed some format checks on one machine, and failed to retrieve rocBLAS package on sles, but passed everything else, so merging.

PR #234, opened 2020-06-11
Adding async functions

daineAMD:
[SWDEV-232444](http://ontrack-internal.amd.com/browse/SWDEV-232444).

Adds:
- hipblasGetVectorAsync
- hipblasSetVectorAsync
- hipblasGetMatrixAsync
- hipblasSetMatrixAsync

Also adds Fortran tests for Get/Set Vector/Matrix functions.

I see some format updates were added here too.


daineAMD:
Failed on ixt-rack-46 but ran on ixt-hq-88 myself and it passed, so it should be ok.

PR #239, opened 2020-07-29
Add static lib build to CI

eidenyoshida:


eidenyoshida:
Static libs portion is passing. Shared libs is not passing but that seems to be a bug not related to this PR

PR #243, opened 2020-08-05
Merge master into develop for ROCm 3.8 feature complete

daineAMD:


daineAMD:
@jzuniga-amd we're seeing some getri failures in hipBLAS. I noticed you recently made some changes, are these failures new or are these related to the ones @cgmb has been looking at?

cgmb:
It's possible it's related to the bug I'm investigating. Sometimes it shows up as a crash, but other times it's just test failures. It's quite consistently the same behaviour until I change my build flags or pull in new changes, so my theory is that there's a (deterministic) out-of-bounds write happening somewhere. Whether it causes a crash or whether it causes incorrect results presumably depends on what's in the region of memory that it ends up writing to.

Do the tests fail when you run them individually?

cgmb:
That said, those look like very specific failures in a piece of functionality that changed quite recently. While it could be related to my bug, my first guess would be that it's a different problem.

daineAMD:
I haven't ran the tests locally, but if it looks like these tests are consistently failing, so probably unrelated to yours.
Troy is looking into this, I'll merge the PRs for now since this is probably a solver issue. We can always add a hotfix later if needed.
Static libs CI is failing as well but looks like that's a CI issue and not related to hipBLAS.

PR #253, opened 2020-09-12
 Make device memory reallocate on demand in hipBLAS

leekillough:
This needs to pick up the correct rocBLAS and rocSOLVER before the CI tests will work.


amcamd (approved):
LGTM as long as CI tests pass. (may need to wait for rocBLAS-internal develop branch artifacts to update.

PR #254, opened 2020-09-14
default fortran compiler to match rocBLAS and rocSOLVER

TorreZuk:
* Fortran compiler consistent with rocBLAS

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,10 @@ if( NOT DEFINED CMAKE_CONFIGURATION_TYPES AND NOT DEFINED CMAKE_BUILD_TYPE )
   set( CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel." )
 endif()
 
+if ( NOT DEFINED CMAKE_Fortran_COMPILER AND NOT DEFINED ENV{FC} )
+  set( CMAKE_Fortran_COMPILER  "gfortran" )
+endif()
+
 project( hipblas LANGUAGES CXX Fortran )
 
 # This finds the rocm-cmake project, and installs it if not found


daineAMD (approved):

cgmb:
This is the same as what we do in rocSOLVER. That ticket was helpful in understanding why. Thanks!

ROCm-cmake warns us that this sort of thing should be handled with a toolchain file. I assume that switching to using a toolchain is planned eventually?

TorreZuk:
I don't see the Fortran compiler rocm cmake warning but yes could later be put in a toolchain file.  For now just matching other libs.  Also build some dependencies outside of cmake so use of the command line override form CXX= is common.

@daineAMD  is this multi-compiler setup all finalized ?

cgmb:
Huh. You're right. I don't see a warning when building rocBLAS, but I see this every time I build rocSOLVER:

```
-- The Fortran compiler identification is GNU 7.5.0

*******************************************************************************
*----------------------------------- ERROR -----------------------------------*
* The variable 'CMAKE_Fortran_COMPILER' should only be set by the cmake toolchain,
* either by calling 'cmake -DCMAKE_Fortran_COMPILER="/usr/bin/gfortran"' or
* set in a toolchain file and added with 
* 'cmake -DCMAKE_TOOLCHAIN_FILE=<toolchain-file>'.
*-----------------------------------------------------------------------------*
*******************************************************************************

CMake Warning at /opt/rocm/share/rocm/cmake/ROCMChecks.cmake:37 (message):
  The toolchain variable 'CMAKE_Fortran_COMPILER' is modified in the
  CMakeLists.txt.
Call Stack (most recent call first):
  build/release/CMakeFiles/3.10.2/CMakeFortranCompiler.cmake:9999 (rocm_check_toolchain_var)
  build/release/CMakeFiles/3.10.2/CMakeFortranCompiler.cmake:1 (set)
  rocsolver/clients/CMakeLists.txt:15 (project)
```

I think it's because rocBLAS has Fortran in its project languages, while rocSOLVER uses enable_language in the clients. So, rocSOLVER ends up setting the variable every time, while rocBLAS normally already has it defined.

Not that any of this is terribly important. I'm just working my way through the dusty corners of our CMake. Eventually, I'll understand the whole process...

PR #256, opened 2020-09-16
ignore cuda if not passed --cuda

TorreZuk:
* don't try and find cuda if not using it

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -449,6 +449,11 @@ pushd .
     cmake_common_options="${cmake_common_options} -DCMAKE_BUILD_TYPE=Debug"
   fi
 
+  # cuda
+  if [[ "${build_cuda}" == false ]]; then
+    cmake_common_options="${cmake_common_options} -DTRY_CUDA=OFF"
+  fi
+
   # clients
   if [[ "${build_clients}" == true ]]; then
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -449,6 +449,11 @@ pushd .
     cmake_common_options="${cmake_common_options} -DCMAKE_BUILD_TYPE=Debug"
   fi
 
+  # cuda
+  if [[ "${build_cuda}" == false ]]; then
+    cmake_common_options="${cmake_common_options} -DTRY_CUDA=OFF"
```
saadrahim, line 454: rocsparse has a flag as well. it is called use-cuda. Should we make it consistent.

TorreZuk: If others agree I can change it, this is just fixing my ability to build without cuda when cuda is partially present.
rocBLAS also is install.sh with --use-cuda, and -DUSE_CUDA.

TorreZuk: Can someone look at the jenkins ?   I still see this CI build error: mv: 'hipblas-0.37.0.434-PR-256-1-4ab1946-ubuntu-18.04_amd64.deb' and 'hipblas-0.37.0.434-PR-256-1-4ab1946-ubuntu-18.04_amd64.deb' are the same file
script returned exit code 1

daineAMD (approved):

PR #258, opened 2020-09-17
Missing types master

daineAMD:
- Added missing conversion functions for these types in hipblas:
  - `HIPBLAS_R_8I`
  - `HIPBLAS_R_32I`

This enables support for some additional gemmEx functions which we support in rocBLAS (mainly i8/i32 mixed precision).

- Added missing tests for complex gemmEx
- Added missing tests for i8/i32 gemmEx

Most of the changes here are incidental from changing the cblas interface. I tried to make few changes in the testing code since this is going to master, but the gemm testing can definitely be cleanup up quite a bit later.
Also apparently in cublas before v10, gemmEx only supports alpha, beta = 0 or 1 for i8/i32 precision. Also that precision isn't supported in their batched versions.

### DIFF
--- a/clients/common/cblas_interface.cpp
+++ b/clients/common/cblas_interface.cpp
@@ -2086,6 +2086,68 @@ void cblas_gemm<hipblasHalf>(hipblasOperation_t transA,
 }
 
 template <>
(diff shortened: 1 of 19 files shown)


amcamd (approved):
LGTM

PR #262, opened 2020-09-22
Remove setting a different default device memory size

leekillough:
With the size of rocBLAS's default device memory allocation increased to 32 MB, we don't really need to allocate a different default in hipBLAS.


amcamd (approved):
LGTM as long as it passes CI

PR #263, opened 2020-09-22
Remove clang++ from CI multi-compiler job

saadrahim:
Making a PR for @eidenyoshida


TorreZuk (approved):
Okay by me.   Do we need to make a ticket to add support for direct clang++ use?

PR #269, opened 2020-10-14
add hipblas-bench, eg usage with gemv

TorreZuk:
* add bench client following rocblas style
* only implemented testing_gemv* changes as prototype use of hipblas-bench


clients/benchmarks/client.cpp
```diff
@@ -0,0 +1,1311 @@
...
+         "BLAS-2 and BLAS-3: second dimension * leading dimension.")
+
+        ("stride_d",
+         value<hipblas_int>(&arg.stride_d)->default_value(128*128),
+         "Specific stride of strided_batched matrix D, is only applicable to strided batched"
+         "BLAS_EX: second dimension * leading dimension.")
+
+        ("stride_x",
+         value<hipblas_int>(&arg.stride_x)->default_value(128*128),
+         "Specific stride of strided_batched vector x, is only applicable to strided batched"
```
amcamd, line 1103: The default values of m, n, k is 128, the default value of incx is 1. With these defaults, I think the default_value for stride_x should be 128, same for stride_y

TorreZuk: Very good will reduce these.  I copied from rocBLAS so those could likely be checked for reduction as well.

amcamd (approved):
LGTM

clients/include/testing_gemv.hpp
```diff
@@ -115,8 +110,20 @@ hipblasStatus_t testing_gemv(Arguments argus)
...
-    for(int iter = 0; iter < 1; iter++)
+    hipStream_t stream;
+    status = hipblasGetStream(handle, &stream);
+    if(status != HIPBLAS_STATUS_SUCCESS)
     {
+        hipblasDestroy(handle);
+        return status;
+    }
+    int runs = argus.timing ? argus.cold_iters + argus.iters : 1;
+    for(int iter = 0; iter < runs; iter++)
```
daineAMD, line 121: I'm wondering if this works correctly when `argus-timing = true && argus.unit_check = true && argus.cold_iters = 2`?
Since we'll be calculating `y = ax + by`, and then using this updated `y` value for `y = ax + by` again.

daineAMD (approved):
One comment, looks good otherwise I think.

clients/include/testing_gemv_batched.hpp
```diff
@@ -135,22 +135,10 @@ hipblasStatus_t testing_gemvBatched(const Arguments& argus)
     }
 
     /* =====================================================================
-           ROCBLAS
+           HIPBLAS
```
daineAMD, line 138: Oh no, who knows how many times I left this comment in hipBLAS.

daineAMD (approved):
LGTM.

PR #271, opened 2020-10-22
prepare for bench additions

TorreZuk:
* prepares for adding bench code by standardizing


clients/include/testing_geam.hpp
```diff
@@ -21,7 +21,7 @@ using namespace std;
 /* ============================================================================================ */
```
daineAMD, line 21: no `#include "testing_common.hpp"` here?

clients/include/testing_set_get_atomics_mode.hpp
```diff
@@ -19,7 +19,7 @@ using namespace std;
```
daineAMD, line 19: Same here and in set_get_vector and set_get_vector_async

daineAMD (approved):
LGTM

clients/include/testing_set_get_atomics_mode.hpp
TorreZuk: Well if they had few includes I didn't update.  I guess I can change them too, unless they won't get benched like atomics, unless maybe we should bench atomics.

clients/common/hipblas_arguments.cpp
```diff
@@ -16,6 +16,7 @@ std::ostream& operator<<(std::ostream& os, std::pair<T1, T2> p)
 {
     os << p.first << ": ";
     os << p.second;
+    return os;
```
cgmb, line 19: You can turn this sort of omission into a compiler error with the flag `-Werror=return-type`. It has basically no false-positives, but I suppose enabling *any* warnings as errors at all is probably a bad idea for a default in an OSS library. We could probably add an option that the CI opts into, though.

TorreZuk: Yes we don't do that for many warnings, usually fixing any warnings is the best policy.   Leave it to @daineAMD  to decide if he wants to adjust hipBLAS I am fine either way.

PR #273, opened 2020-10-28
Adding gemm and trsm to hipblas-bench

daineAMD:


clients/gtest/trsm_gtest.cpp
```diff
@@ -140,7 +140,7 @@ Arguments setup_trsm_arguments(trsm_tuple tup)
     arg.transA_option = side_uplo_transA_diag[2];
     arg.diag_option   = side_uplo_transA_diag[3];
 
-    arg.timing = 1;
+    arg.timing = 0;
```
TorreZuk, line 143: Not clear how this gets replicated, you want timing on for the bench?

TorreZuk (approved):
LGTM, I guess you can compare against rocblas-bench to see if you have all the parameters logged.   But rocblas-bench might be the one missing some!  Do the times look the same vs. rocblas-bench?

TorreZuk:
The CI errors I don't understand as develop seemed to build.  If you think it could be my cmake cleanup PR let me know if you want me to help investigate, but how would develop be all passing?

daineAMD:
I re-ran develop and it seems to be failing identically to this PR, so should be unrelated to any hipBLAS changes.

clients/gtest/trsm_gtest.cpp
daineAMD: This is turning timing off for hipblas-test, I'm not sure why it was set to 1 here to begin with, especially since it didn't do anything previously.

daineAMD:
Timing is similar between rocBLAS/hipBLAS (also with cuBLAS/hipBLAS), only a few microseconds different on average. As mentioned, we are reporting the time to run all iterations rather than the average. I'll open a new pull request today to fix that so it's consistent across rocBLAS and hipBLAS.

PR #276, opened 2020-11-03
Adding cpack solver dependency and updating version req's

daineAMD:
Adds cpack dependency on rocSolver.

Also updating the required version of rocBLAS/rocSOLVER. This requires the rocBLAS version released with rocm 3.8 as that is where the atomic functions were added - which hipBLAS supports. Requires rocSOLVER version 3.7 for getri which hipBLAS supports.


TorreZuk (approved):
Looks good to me, assume you tested as it seems we must not test this for release.  @amcamd should we make a ticket for QA to do package based installation testing?

amcamd:
@TorreZuk , yes, go ahead and make a ticket for QA to do package based installation testing.

amcamd (approved):
LGTM as long as it passes CI testing

daineAMD:
Updated to rocSolver 3.8 since solver 3.7 isn't compatible with rocblas 3.8, seems to be working fine.
As for the ticket, is it going to just be for the latest release, or are we testing that installation works with previous releases as specified here?

amcamd:
For the ticket I think we need to test for the latest release. We assume that when they install hipBLAS it will see a dependency on rocBLAS and rocSOLVER, and it will install the latest rocBLAS and rocSOLVER.

TorreZuk:
> @TorreZuk , yes, go ahead and make a ticket for QA to do package based installation testing.

@saadrahim @eidenyoshida do we have a request to QA for release testing to ensure packages dependencies are correct already?  
If not shall I do that, or can math-CI test the package installation?

library/CMakeLists.txt
```diff
@@ -60,8 +60,8 @@ add_subdirectory( src )
 # endif( )
 
 # Package specific CPACK vars
-set( CPACK_DEBIAN_PACKAGE_DEPENDS "rocblas (>= 0.10.3)" )
-set( CPACK_RPM_PACKAGE_REQUIRES "rocblas >= 0.10.3" )
+set( CPACK_DEBIAN_PACKAGE_DEPENDS "rocblas (>= 2.28.0), rocsolver (>= 3.8.0)" )
+set( CPACK_RPM_PACKAGE_REQUIRES "rocblas >= 2.28.0, rocsolver >= 3.8.0" )
```
saadrahim, line 64: Should this be == instead of >=. We do not support mixing and matching currently.

eidenyoshida:
> > @TorreZuk , yes, go ahead and make a ticket for QA to do package based installation testing.
> 
> @saadrahim @eidenyoshida do we have a request to QA for release testing to ensure packages dependencies are correct already?
> If not shall I do that, or can math-CI test the package installation?

math-ci doesn't check package dependencies as far as I know

library/CMakeLists.txt
daineAMD: How would setting it to == play out? It seems to me that it would cause issues during the development cycle, especially if we're requiring a certain version here on the develop branch. If version `x` gets released tomorrow, would we then want to update this to `== x` immediately? What would it look like when moving to release branches?
I think that documenting that mixing versions isn't supported is enough, I'm not sure it has to be enforced here.

TorreZuk: We won't plan on taking things out of the API so >= to me makes sense.  Hipblas can likely work with newer version of rocsolver, but rocsolver can likely only work with one version of rocBLAS.  It could be we only depend on rocSOLVER as that will pull in rocBLAS, unless it is being built as without solver support  :-(

saadrahim: If rocsolver is tightly coupled with a specific rocblas version, I think we should enforce it in the packaging for rocsolver. hipblas can depend on rocsolver to provide rocblas. This may head off future user complaints. Just my suggestion. Ideally, rocsolver wouldn't be tightly coupled with rocblas, but we should enforce what is true now.

amcamd:
We use semantic versioning with produce version numbers having major.minor.patch.   As long as there is no change in major, we should be backward compatible. We have not had a change in major for over 2 years.  I hope we can keep with >= rather than == . rocBLAS depends on Tensile, and we bump the Tensile version required by rocBLAS at every Feature Complete merge. In hipBLAS I think we will need to bump the rocBLAS and rocSOLVER versions at Feature Complelte.

jzuniga-amd:
I am not sure what is the problem with the failing builds. I tested the same combination of rocBLAS and rocSOLVER commits on my machine and it builds with no problem. Also, rocSOLVER is working on CI with tip of rocBLAS develop. I don't think there is anything to be fix in rocSOLVER for the new GEMV templates.

@eidenyoshida can we re-run the tests for this PR using latest successful builds of rocBLAS and rocSOLVER again?

TorreZuk:
Okay digging in I see it is failing when using an old rocSOLVER build from October on some platforms.  So we are testing different builds on different platforms.   This is a bit confusing but maybe a one off failure?

eidenyoshida:
I just triggered new builds on CI

cgmb:
> Okay digging in I see it is failing when using an old rocSOLVER build from October on some platforms. So we are testing different builds on different platforms. This is a bit confusing but maybe a one off failure?

Perhaps that's because of the rocm-ci crash. IMO, our CI should really be trying to use a 'last known-good' configuration when testing incremental changes, and incrementally advancing that configuration as we discover newer working configs. It might occasionally need a manual poke when there's synchronized changes required across multiple libraries, or when an unstable configuration gets marked as good, but it would prevent so many problems with version mismatches.

> We use semantic versioning with produce version numbers having major.minor.patch. As long as there is no change in major, we should be backward compatible.

How is this normally handled in Debian? Is the major version restriction normally done with an extra set of constraints, like `rocblas (>= 2.28.0), rocblas (< 3.0.0), rocsolver (>= 3.8.0), rocsolver (< 4.0.0)`, or should it be `rocblas2 (>= 2.28.0), rocsolver3 (>= 3.8.0)` with a package name bump on major versions?

eidenyoshida:
For hipBLAS jobs, math-ci will download and install rocSOLVER and rocBLAS packages from the last passing precheckin builds from math-ci (not rocm-ci). Not entirely sure what happened with the last build, but the current CI builds seem to be passing

cgmb:
> For hipBLAS jobs, math-ci will download and install rocSOLVER and rocBLAS packages from the last passing precheckin builds from math-ci (not rocm-ci).

If I understand our current system correctly, that's fundamentally different from what I'm suggesting. The CI currently downloads the most recent version of rocBLAS that has passed the rocBLAS tests, and the most version of rocSOLVER that has passed the rocSOLVER tests (based on whatever happened to be the latest rocBLAS version at the time when those tests were run). It doesn't necessarily use versions that were known to work with hipBLAS.

There are some downsides to what I'm suggesting, but it would definitely be more stable for PRs. We could probably reserve the more aggressive integration tests for the branch tips.

eidenyoshida (approved):

daineAMD:
Updates in the latest commit:
- updated dependencies to latest versions in master branch (corresponding to 3.10).
- added updating versions to bump_* scripts.
    - develop branch of hipblas will depend on version on latest develop branch of rocblas (current is 2.33.0, will be updated to 2.35.0 for next release)
    - master branch of hipblas will depend on version on latest master branch of rocblas (will be updated to 2.34.0 for next release)
    - both hipblas branches currently depend on rocsolver 3.10.0, will be updated to 4.0.0 for next release

amcamd:
LGTM

PR #277, opened 2020-11-05
Adding performance measurement scripts

daineAMD:
Allows performance measurements as done in rocBLAS to be done in hipBLAS as well.
Essentially the files added in this PR are directly copied from rocBLAS and changed from "rocBLAS" to "hipBLAS".

An aside: The testing in hipBLAS is currently done differently than rocBLAS, as it doesn't use *_gentest.py and/or yaml files, but this is also a step in that direction if we ever decide to change it, so I didn't remove any currently unused code here. Also, hipblas_gentest.py and hipblas_common.yaml have been added to the sctips/* directory rather than the clients/* directory (as it is in rocBLAS), so I didn't add any dependencies for yaml. These would be moved if we ever move to the rocBLAS testing structure.

If we ever go for some sort of "utilities" repo, this is another good candidate to be made generic and shared between some libraries.


TorreZuk (approved):
LGTM , yes common place would have been nice.   Too bad we don't have generic types across more components.  e.g. rocm_int64, rocm_complex,  then all these type maps could be kept the same.

daineAMD:
Seems to work on on AMD hardware, but does not yet support CUDA hardware.

These scripts won't be released, and isn't documented, so unless I hear otherwise I will go ahead and merge this PR later today. A future PR will add support for CUDA - I think this will make the next PR more readable/reviewable which may be important since it might have some more interesting additions rather than a simple copy from rocBLAS.

PR #283, opened 2020-11-19
Removing rocm dependency for CUDA builds

daineAMD:
After removing all rocm libraries, and installing hip-nvcc, this seems to work on a CUDA machine, tests are currently running. This seems to be consistent with hipSPARSE for the most part. @YvanMokwinski if you notice anything that drastically differs, let me know!

Sidenote: The dependency installation seems to be weird on my CUDA machine, I had to do a small workaround to get them to be found, I'll look into this tomorrow, but it's unrelated to this PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,9 +52,15 @@ include( ROCMInstallSymlinks )
 set ( VERSION_STRING "0.41.0" )
 rocm_setup_version( VERSION ${VERSION_STRING} )
 
+if( NOT DEFINED $ENV{HIP_PATH})
+    set( HIP_PATH "/opt/rocm-3.9.0/hip" )
+else( )
+    set (HIP_PATH $ENV{HIP_PATH} )
+endif( )
+
 # Append our library helper cmake path and the cmake path for hip (for convenience)
 # Users may override HIP path by specifying their own in CMAKE_MODULE_PATH
-list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake )
+list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake  ${HIP_PATH}/cmake)
 
 # NOTE:  workaround until hip cmake modules fixes symlink logic in their config files; remove when fixed
 list( APPEND CMAKE_PREFIX_PATH /opt/rocm /opt/rocm/llvm /opt/rocm/hip )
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -72,7 +78,7 @@ option( BUILD_SHARED_LIBS "Build hipBLAS as a shared library" ON )
 
 
 # Hip headers required of all clients; clients use hip to allocate device memory
-find_package( hip REQUIRED CONFIG PATHS ${ROCM_PATH} /opt/rocm)
+find_package( HIP REQUIRED )
```
TorreZuk, line 82: Maybe just try
if (USE_CUDA) 
find_package HIP 
else()
find_package( hip REQUIRED CONFIG PATHS ${ROCM_PATH} /opt/rocm)

and then the romc builds are the same as hip provides the hip::host
One of the hip dependencies was the library I already removed (issues reported in 3.9), so ensuring all hip references are not in the if USE_CUDA may suffice.

library/src/CMakeLists.txt
```diff
@@ -121,12 +121,20 @@ rocm_install_targets(
...
-  TARGETS roc::hipblas
-  PREFIX hipblas
-  DEPENDS PACKAGE HIP
-  NAMESPACE roc::
- )
+if ( NOT USE_CUDA )
+    rocm_export_targets(
+        TARGETS roc::hipblas
+	PREFIX hipblas
+	DEPENDS PACKAGE HIP
```
TorreZuk, line 128: I think this should now be hip not HIP.

TorreZuk: And maybe the CUDA one depends on HIP ?

TorreZuk (approved):

haampie:
@TorreZuk, this PR doesn't really make sense to me...

For one, `find_package(hip)` should be searching for the hip config file (with hip::device/hip::host/hip::amd64something targets) and `find_package(HIP)` should be looking for the HIP module file (for hip_add_library macro's).

If this is intended, can you please make it use `find_package(HIP MODULE)` and `find_package(hip CONFIG)`, because otherwise find_package(HIP) will just find the config file if that's in the CMAKE_PREFIX_PATH.

Better would be to just rename FindHIP.cmake to FindHIPTools.cmake in HIP, but I'm afraid that ship has sailed.

haampie:
See #295

TorreZuk:
The find_package should not ignore case of the PackageName in either module or config mode from what I have seen,  are you finding that it ignored the case and used the **hip** config when no **HIP** could be found?  Yes agree they should've renamed the FindHIP long ago.  This PR removes the use of "hip" when USE_CUDA is on (USE_CUDA is false by default in install.sh but as you saw the cmakelists.txt default was ON) so this PR makes sense to me, but as your #295 I agree that looks fine and will review it there.

Relevant cmake doc:
`If the MODULE option is not specified in the above signature, CMake first searches for the package using Module mode. Then, if the package is not found, it searches again using Config mode. A user may set the variable CMAKE_FIND_PACKAGE_PREFER_CONFIG to TRUE to direct CMake first search using Config mode before falling back to Module mode.`

PR #285, opened 2020-11-27
Fixed erroneous caxpy and zaxpy method signatures

tfalders:
Related to https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/188

The method signature for caxpy in norm.cpp had alpha as a float, whereas the LAPACK signature has alpha as a single complex. A similar error affects the norm.cpp version of zaxpy.

A separate fix has dockerfile-build-nvidia-cuda point to hipBLAS directories, rather than rocBLAS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log for hipBLAS
 
+## [(Unreleased) hipBLAS 0.42.0 for ROCm 4.1.0]
+### Fixed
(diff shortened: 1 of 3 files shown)


tfalders:
This appears to be the source of the complex test failures I encountered here: https://github.com/ROCmSoftwarePlatform/hipBLAS/pull/222#issuecomment-635647971

daineAMD (approved):
Glad you were able to find these errors!

PR #286, opened 2020-12-01
Adding gtest dependency to bench

daineAMD:


TorreZuk (approved):
LGTM, not sure how the non-cuda build didn't need it?   Maybe higher level/common check?

daineAMD:
I'm not sure either. On my local hip machine there happened to be a copy of gtest in /usr/local when I was building, not sure if that happened to be the case for all hip machines.

PR #287, opened 2020-12-02
Adding support for blas1 _ex functions

daineAMD:
Adding support for:
- axpy_ex
- dot(c)_ex
- nrm2_ex
- rot_ex
- scal_ex

Some precisions are supported in rocBLAS and cuBLAS, so if the tests do not return success, we check that they return not implemented.

The CUDA documentation for some functions (such as [axpy](https://docs.nvidia.com/cuda/cublas/index.html#cublas-axpyEx)) claim support  for mixed f16_r/f32_r precisions. The documentation does not explicitly specify which type alpha is in this case. Based on the wording of the documentation, I added support for alpha=f16_r in these cases, but it appears that isn't the case as these return not_supported in CUDA. Because of this, I'll likely be adding support for alpha=f32_r in rocBLAS for these in the future since this is probably what is supported in CUDA.


clients/gtest/dot_ex_gtest.cpp
```diff
@@ -0,0 +1,265 @@
...
+    // if not success, then the input argument is problematic, so detect the error message
+    if(status != HIPBLAS_STATUS_SUCCESS)
+    {
+        if(arg.N < 0 || arg.incx < 0 || arg.incy < 0)
+        {
+            EXPECT_EQ(HIPBLAS_STATUS_INVALID_VALUE, status);
+        }
+        else
+        {
+            EXPECT_EQ(HIPBLAS_STATUS_NOT_SUPPORTED, status); // for cuda
```
TorreZuk, line 138: Can we log these unsupported tests to keep track of them?

clients/include/testing_nrm2_strided_batched_ex.hpp
```diff
@@ -0,0 +1,184 @@
...
+
+#include "testing_common.hpp"
+
+using namespace std;
+
+/* ============================================================================================ */
+
+// Tolerance of 100 fails for complex,
+// TODO: something better than arbitrary tolerance.
+template <typename T>
```
TorreZuk, line 18: Is there a ticket made for all these TODOs?

library/include/hipblas.h
```diff
@@ -7103,6 +7149,264 @@ HIPBLAS_EXPORT hipblasStatus_t hipblasTrsmStridedBatchedEx(hipblasHandle_t    ha
...
+                                                    hipblasDatatype_t executionType);
+
+HIPBLAS_EXPORT hipblasStatus_t hipblasAxpyStridedBatchedEx(hipblasHandle_t   handle,
+                                                           int               n,
+                                                           const void*       alpha,
+                                                           hipblasDatatype_t alphaType,
+                                                           const void*       x,
+                                                           hipblasDatatype_t xType,
+                                                           int               incx,
+                                                           int               stridex,
```
TorreZuk, line 7215: Aren't these strides to all be int64, some cuda seem to be long long int .... ?

TorreZuk (approved):
LGTM if tests pass.  A lot of work in this PR
Wwould adding the batched_host vector be worth it for the testing?   Could be a ticket should be made.

library/include/hipblas.h
daineAMD: Yeah, strides are currently `int` across hipBLAS and need to be updated to at least match rocBLAS, I'll make a ticket.

clients/include/testing_nrm2_strided_batched_ex.hpp
daineAMD: Not yet, on it!

clients/gtest/dot_ex_gtest.cpp
daineAMD: Hm, similar to the  skipping tests aspect done in rocBLAS, you mean?

TorreZuk: Just so we can count how many not supported by grepping the log, something output.  Not really skipped

daineAMD:
[host_batched_vector ticket](http://ontrack-internal.amd.com/browse/LWPMLSE-96).
[nrm2 testing TODOs ticket](http://ontrack-internal.amd.com/browse/LWPMLSE-95).
[update hipBLAS strides to int64_t ticket](http://ontrack-internal.amd.com/browse/LWPMLSE-94).

The last 2 shouldn't take long at all, so I'll add those soon.

PR #288, opened 2020-12-07
Change stride from int to int64_t

daineAMD:
Most changes from a search/replace, steps completed here are:

- Add `typedef int64_t hipblasStride`.
- Search and replace for `(int)(\s)+(stride)` -> `hipblasStride stride`.
- Small fixes after for a few missed (in clients/benchmarks/client.cpp) and extra (stride_scale) replacements.

Builds OK on my hip machine, tests currently running on a CUDA machine. Will rely on CI for hip tests.


TorreZuk (approved):
LGTM if tests pass.  Looks like you covered almost all the calculation promotion with your double strideScale but saw a couple that might need casts added but can wait until large test sizes show a need for such.

PR #289, opened 2020-12-08
Updating nrm2 tolerance code

daineAMD:
Matching rocBLAS tolerance checking for nrm2.
See discussion [here](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/366#discussion_r504300139).

### DIFF
--- a/clients/include/unit.h
+++ b/clients/include/unit.h
@@ -53,11 +53,11 @@ void unit_check_error(T error, T tolerance)
 }
 
 template <typename T>
-void unit_check_nrm2(T cpu_result, T gpu_result, T tolerance)
+void unit_check_nrm2(T cpu_result, T gpu_result, int vector_length)
 {
-    T allowable_error = tolerance * std::numeric_limits<T>::epsilon() * cpu_result;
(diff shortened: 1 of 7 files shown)


clients/include/unit.h
```diff
@@ -63,4 +63,21 @@ void unit_check_nrm2(T cpu_result, T gpu_result, T tolerance)
 #endif
 }
 
+template <typename T>
+void unit_check_nrm2(int            batch_count,
+                     host_vector<T> cpu_result,
+                     host_vector<T> gpu_result,
+                     T              tolerance)
```
TorreZuk, line 70: Why not rename the last arg vector_length and just pass in vector length N?  Tolerance implies something to be judged which seems large as 2N.   You can type 2*N once, as later you can also tighten in one place, sqrt(2N),... 2sqrt(N) ?

TorreZuk (approved):
Suggestions added

clients/include/unit.h
daineAMD: Renamed to vector_length and am now passing in N. Also removed scalar 2 as it looks like all tests pass without it (on cuda and hip).

TorreZuk: Very good it is already paying off ;-)

PR #290, opened 2020-12-09
Adding updated batched vectors for tests - only on scal

daineAMD:
This adds the `host_batch_vector` and `device_batch_vector` implementations from rocBLAS. I've updated the implementation of scal_batched tests to the intended use. I will update all other tests in another PR to keep this one moderately small and easy to review.


TorreZuk (approved):
LGTM

PR #291, opened 2020-12-15
Add features to performance scripts

daineAMD:
This adds the following features to the hipBLAS performance scripts:
- Allows performance scripts to be ran on CUDA machines.
- Allows the creation of efficiency plots (this takes GFLOPS and compares it to the theoretical max for GFLOPS)
- Allows 2 runs to be compared on the same plot. Currently this is solely to compare a single run to an already-completed run from a CUDA machine. This can be expanded in the future if people are interested in comparing other data.

I've also opened [this ticket](http://ontrack-internal.amd.com/browse/LWPMLSE-102) to clean up some of the performance plotting code.

Note that I'm the only one that's currently using these scripts and that the changes to them are ongoing, they still need a lot of improvement, but these changes currently work for my needs.

### DIFF
--- a/scripts/performance/blas/commandrunner.py
+++ b/scripts/performance/blas/commandrunner.py
@@ -708,6 +708,9 @@ class RunConfiguration(object):
     def _machine_specs_filename(self):
         return os.path.join(self.output_directory, "specs.json")
 
+    def _machine_specs_filanem_compare(self):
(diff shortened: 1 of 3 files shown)


TorreZuk (approved):
If it works for you it works for me.  Not sure how it compares to the other scripts but leave that for others to comment on.

PR #292, opened 2020-12-17
Add CUDA CI jenkinsfile

eidenyoshida:


eidenyoshida:
I've added a new job called 'Precheckin-CUDA' that you can see in the github status checks. Right now it's running on our bleeding-edge job. It won't be on the main job until we advance the build number.

daineAMD (approved):
Awesome! Thanks Eiden!

PR #294, opened 2021-01-08
Fixing half-precision alpha for gemm tests.

daineAMD:
Previously wasn't handling this correctly, the simple cast doesn't work as expected. This matches gemm_ex tests for now. Also it doesn't look like we have tests for a non-zero imaginary part for cgemm/zgemm, I'll make a ticket for this.

### DIFF
--- a/clients/include/testing_gemm_strided_batched.hpp
+++ b/clients/include/testing_gemm_strided_batched.hpp
@@ -1,5 +1,5 @@
 /* ************************************************************************
- * Copyright 2016-2020 Advanced Micro Devices, Inc.
+ * Copyright 2016-2021 Advanced Micro Devices, Inc.
  *
  * ************************************************************************ */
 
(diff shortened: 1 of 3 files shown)


clients/include/testing_gemm_strided_batched.hpp
```diff
@@ -44,8 +44,22 @@ hipblasStatus_t testing_gemm_strided_batched(const Arguments& argus)
...
+    if(is_same<T, hipblasHalf>::value)
+    {
+        alpha = float_to_half(alpha_float);
+        beta  = float_to_half(beta_float);
+    }
+    else
+    {
+        alpha = static_cast<T>(alpha_float);
+        beta  = static_cast<T>(beta_float);
+    }
```
TorreZuk, line 62: Not for this release but expect this should move to use helpers the same as rocblas.  If you can get_alpha<T> and get_beta<T> you can handle this with specializations along with NaN support.

TorreZuk (approved):
LGTM if the tests pass

daineAMD:
Passed everyone save for the rocSOLVER centos CI issue - merging.

PR #295, opened 2021-01-15
Don't look for CUDA by default

haampie:
Detection of whether HIP uses HCC or NVCC should really be improved, but I leave that up to AMD.

How it's done here with `USE_CUDA=ON` by default, and then

```
if( NOT CUDA_FOUND )
  target_compile_definitions( hipblas-bench PRIVATE __HIP_PLATFORM_HCC__ )
```

is really poor practice, cause you can have a system CUDA install but still have HIP configured with HCC.

Please just infer from the HIP config file if the compiler is HCC or NVCC, don't assume "if the user has CUDA they likely want CUDA".

Thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -87,7 +87,7 @@ endif()
 if( USE_CUDA)
     find_package( HIP REQUIRED MODULE )
 else( )
-    find_package(hip REQUIRED CONFIG PATHS ${ROCM_PATH} /opt/rocm)
+    find_package( hip REQUIRED CONFIG PATHS ${ROCM_PATH} /opt/rocm)
 endif( )
 
 if( USE_CUDA )
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,8 +79,8 @@ option( BUILD_SHARED_LIBS "Build hipBLAS as a shared library" ON )
 # Quietly look for CUDA, but if not found it's not an error
 # The presense of hip is not sufficient to determine if we want a rocm or cuda backend
 option(USE_CUDA "Look for CUDA and use that as a backend if found" OFF)
-if(USE_CUDA)
-    find_package( CUDA QUIET )
+if (USE_CUDA)
+    find_package( CUDA REQUIRED )
 endif()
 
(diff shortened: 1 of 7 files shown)


TorreZuk (approved):
Making the default USE_CUDA off to match the install.sh coding makes sense to me, the CUDA CI may require adjustment but that is still work in progress.   Would mainly affect the failure modes and all fine by me, as if you say USE_CUDA we don't really need a fallback if there is no CUDA.

CMakeLists.txt
```diff
@@ -76,18 +76,17 @@ endif( )
 # BUILD_SHARED_LIBS is a cmake built-in; we make it an explicit option such that it shows in cmake-gui
 option( BUILD_SHARED_LIBS "Build hipBLAS as a shared library" ON )
 
-# Quietly look for CUDA, but if not found it's not an error
-# The presense of hip is not sufficient to determine if we want a rocm or cuda backend
-option(USE_CUDA "Look for CUDA and use that as a backend if found" ON)
-if(USE_CUDA)
-    find_package( CUDA QUIET )
+# Find CUDA if the user wants a CUDA version.
+option(USE_CUDA "Look for CUDA and use that as a backend if found" OFF)
```
daineAMD, line 80: Maybe I'm mistaken, but does this change not make `USE_CUDA` always OFF? In install.sh we turn it off when the `--cuda` flag is not used, so it's never on.
This failed on a CUDA system for me - it was using the incorrect (rocblas) header file.

TorreZuk: --cuda in the install.sh script would have to turn in ON now.   The change states OFF will be the default value unless a -DUSE_CUDA=ON which is passed by install.sh, others may not use the install.sh or may use cmake-gui to set options.   You can add a commit for this ?

haampie: I've pushed the change, didn't realize there was a shell script!

daineAMD (approved):
Seems to build fine on CUDA now! @haampie I will go ahead and merge this after CI passes (in ~an hour) if you're ok with that.
Thanks so much for the contribution!

haampie:
Thank you! Very happy that you're accepting PRs here :)

PR #299, opened 2021-01-19
Adding support for gemm_ex in hipblas-bench

daineAMD:


clients/common/norm.cpp
```diff
@@ -171,6 +171,27 @@ double norm_check_general<hipblasBfloat16>(
...
+    host_vector<float> hGPU_double(N * lda);
+
+    for(int i = 0; i < M; i++)
+    {
+        for(int j = 0; j < N; j++)
+        {
+            hCPU_double[i + j * lda] = (hCPU[i + j * lda]);
+            hGPU_double[i + j * lda] = (hGPU[i + j * lda]);
+        }
+    }
```
NaveenElumalaiAMD, line 190: Does it have to be `hCPU_float`, `hGPU_float` instead of `hCPU_double`, `hGPU_double`?

NaveenElumalaiAMD (approved):

clients/common/norm.cpp
daineAMD: Yep! Fixed.

PR #304, opened 2021-02-08
Added bandwidth calculation and SWAP

NaveenElumalaiAMD:
Added support for bandwidth based performance graphs to hipBLAS performance script which can be used for certain level 1 functions and added Level: -1 SWAP to hipblas-bench


daineAMD (approved):
Looks fine as long as the new bandwidth functionality works for you!

PR #306, opened 2021-02-23
Updating int8 gemm_ex

daineAMD:
Updating hipBLAS to work correctly with the new rocBLAS gemm_ex.

- When calling hipblasGemmEx with a rocBLAS backend, we now query the preferable data format for the architecture. This preferred data format is then passed into the call for rocblas_gemm_ex.
- Slight additions were needed to be able to query the architecture from hipBLAS clients to get our tests to work. Note I haven't tested this on gfx908 yet, will wait for CI.

The changes to the testing infrastructure (that is, adding the architecture query) is needed as we are not exposing the rocblas_query API function in hipBLAS with this PR. Any user that is building a program to run on multiple architectures will have to do something similar, unless we decide to expose the query API in hipBLAS as well.

### DIFF
--- a/library/src/hcc_detail/hipblas.cpp
+++ b/library/src/hcc_detail/hipblas.cpp
@@ -14239,8 +14239,9 @@ hipblasStatus_t hipblasZgemmStridedBatched(hipblasHandle_t             handle,
 }
 
 // gemm_ex
-// First, we are querying the intended int8 layout from rocBLAS and passing in the result.
-// Users should ensure that they are passing in the correct data layout for this function.
+// Note for int8 users - For rocBLAS backend, please read rocblas_gemm_ex documentation on int8
+// data layout requirements. hipBLAS makes the assumption that the data layout is in the preferred
+// format for a given device as documented in rocBLAS.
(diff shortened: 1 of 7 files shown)


library/src/hcc_detail/hipblas.cpp
```diff
@@ -14239,6 +14239,8 @@ hipblasStatus_t hipblasZgemmStridedBatched(hipblasHandle_t             handle,
 }
 
 // gemm_ex
+// First, we are querying the intended int8 layout from rocBLAS and passing in the result.
```
bragadeesh, line 14242: This comment needs some cleanup to improve clarity.
Can you change: "Note for int8 users - please read rocblas_gemm_ex documentation on int8 data layout requirements. hipBLAS makes the assumption that the data layout is in the preferred format for a given device as documented in rocBLAS."

bragadeesh:
LGTM as long as CI and tests pass

amcamd (approved):

daineAMD:
Single CI failure seems to be unrelated, merging.

PR #309, opened 2021-02-27
ROCm 4.2 merge staging into develop

daineAMD:


amcamd (approved):
LGTM as long as tests pass

daineAMD:
Seeing some segmentation faults on the first test in some cases. I believe this is non-hipBLAS related, but I'm going to check out the failing docker containers and investigate.
I think it's safe to merge this as it's failing the same way on develop already, and all the tests have passed in other configs.

PR #310, opened 2021-03-03
Add scal_ex to hipblas-bench

NaveenElumalaiAMD:

### DIFF
--- a/clients/include/testing_scal_batched_ex.hpp
+++ b/clients/include/testing_scal_batched_ex.hpp
@@ -14,7 +14,7 @@ using namespace std;
 /* ============================================================================================ */
 
 template <typename Ta, typename Tx = Ta, typename Tex = Tx>
-hipblasStatus_t testing_scal_batched_ex_template(Arguments argus)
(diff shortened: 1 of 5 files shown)


daineAMD (approved):

clients/include/testing_scal_batched_ex.hpp
```diff
@@ -109,20 +112,53 @@ hipblasStatus_t testing_scal_batched_ex_template(Arguments argus)
...
+        }
+        gpu_time_used = get_time_us_sync(stream) - gpu_time_used;
+
+        ArgumentModel<e_N, e_alpha, e_incx, e_batch_count>{}.log_args<Tx>(
+            std::cout,
+            argus,
+            gpu_time_used,
+            scal_gflop_count<Tx, Ta>(N),
+            scal_gbyte_count<Tx>(N),
+            hipblas_error);
```
daineAMD, line 155: I noticed that we don't calculate `hipblas_error` anywhere. This should be calculated using `norm_check_general` in the `if(unit_check)` section. I notice it's missing in `testing_scal.hpp` as well, might be missing in other places as well but that can be a different PR.

PR #311, opened 2021-03-03
Add trmm to hipblas-bench

daineAMD:


clients/include/testing_trmm_strided_batched.hpp
```diff
@@ -136,6 +132,66 @@ hipblasStatus_t testing_trmm_strided_batched(const Arguments& argus)
         {
             unit_check_general<T>(M, N, batch_count, ldb, stride_B, hB, hB_copy);
         }
+        hipblas_error = norm_check_general<T>(
+            'F', M, N, ldb, stride_B, hB_copy.data(), hB.data(), batch_count);
```
TorreZuk, line 136: Minor question so does -v sets both unit and arge?   If unit_check is sometimes without norm_check would you want to if (argus.norm_check) to avoid this for test runs?

TorreZuk (approved):
LGTM with minor question asked.

clients/include/testing_trmm_strided_batched.hpp
daineAMD: I'm pretty sure unit_check is only used for tests, and norm_check is only used for bench. I'll check tomorrow, if that's the case I'll add an if statement.

clients/include/flops.hpp
```diff
@@ -720,42 +720,21 @@ constexpr double syrkx_gflop_count<hipblasDoubleComplex>(int n, int k)
...
-    if(HIPBLAS_SIDE_LEFT == side)
-    {
-        return (1.0 * m * n * (m + 1)) / 1e9;
-    }
-    else
-    {
-        return (1.0 * m * n * (n + 1)) / 1e9;
-    }
+    return (1.0 * m * n * (k + 1)) / 1e9;
 }
```
amcamd, line 726: I think it is best to get the flop count from the value NOP from the following file http://www.netlib.org/blas/sblas3time.f contains:

            CALL STRMM( SIDE, UPLO, TRANSA, 'N', M, N, ALPHA, A, LDA, B,
     $                  LDB )
            S2 = SECOND( )
            TIME = ( S2 - S1 ) - ( S1 - S0 )
            IF( SIDE.EQ.'L' )THEN
               NOP = M*M*N
            ELSE
               NOP = M*N*N
            END IF

The number of operations for trmm is m*m*n if HIPBLAS_SIDE_LEFT == side, and m*n*n if HIPBLAS_SIDE_LEFT != side

clients/include/flops.hpp
```diff
@@ -720,42 +720,21 @@ constexpr double syrkx_gflop_count<hipblasDoubleComplex>(int n, int k)
...
-    if(HIPBLAS_SIDE_LEFT == side)
-    {
-        return 4 * (1.0 * m * n * (m + 1)) / 1e9;
-    }
-    else
-    {
-        return 4 * (1.0 * m * n * (n + 1)) / 1e9;
-    }
+    return 4 * (1.0 * m * n * (k + 1)) / 1e9;
 }
```
amcamd, line 732: similar to comment above, I think it is best to get trmm_gflop_count from the following code from the file http://www.netlib.org/blas/cblas3time.f 

            S1 = SECOND( )
            CALL CTRMM( SIDE, UPLO, TRANSA, 'N', M, N, ALPHA, A, LDA, B,
     $                  LDB )
            S2 = SECOND( )
            TIME = ( S2 - S1 ) - ( S1 - S0 )
            IF( SIDE.EQ.'L' )THEN
               NOP = 4*M*M*N
            ELSE
               NOP = 4*M*N*N
            END IF
            IF( TIME.LE.0.0E0 )THEN
               RESLTS( IMVAL, INVAL ) = 0.0E0
            ELSE
               RESLTS( IMVAL, INVAL ) = NOP/( 1.0E6*TIME )
            END IF

for rocBLAS the value would be 4*m*m*n/1e9 if HIPBLAS_SIDE_LEFT == side and 4*m*n*n if HIPBLAS_SIDE_LEFT != side

clients/include/flops.hpp
```diff
@@ -720,42 +720,21 @@ constexpr double syrkx_gflop_count<hipblasDoubleComplex>(int n, int k)
...
-    if(HIPBLAS_SIDE_LEFT == side)
-    {
-        return (1.0 * m * n * (m + 1)) / 1e9;
-    }
-    else
-    {
-        return (1.0 * m * n * (n + 1)) / 1e9;
-    }
+    return 4 * (1.0 * m * n * (k + 1)) / 1e9;
 }
```
amcamd, line 738: see comment above on trmm_gflop_count<hipblasComplex>

clients/include/testing_trmm.hpp
```diff
@@ -99,6 +99,53 @@ hipblasStatus_t testing_trmm(const Arguments& argus)
...
+                      e_M,
+                      e_N,
+                      e_lda,
+                      e_ldb>{}
+            .log_args<T>(std::cout,
+                         argus,
+                         gpu_time_used,
+                         trmm_gflop_count<T>(M, N, K),
+                         trmm_gbyte_count<T>(M, N, K),
+                         hipblas_error);
```
amcamd, line 148: this will be    trmm_gflop_count<T>(M,N,side)

clients/include/bytes.hpp
```diff
@@ -199,6 +199,13 @@ constexpr double gemm_gbyte_count(int m, int n, int k)
     return (sizeof(T) * (m * k + n * k + m * n)) / 1e9;
 }
 
+/* \brief byte counts of TRMM */
+template <typename T>
+constexpr double trmm_gbyte_count(int m, int n, int k)
+{
+    return (sizeof(T) * (m * n + tri_count(k))) / 1e9;
+}
+
```
amcamd, line 208: I think this will need to change to:
if(HIPBLAS_SIDE_LEFT == side)
    return (sizeof(T) * (m*n*2 + m*m/2))/1e9
else
   return (sizeof(T) * (m*n*2 + n*n/2))/1e9

This is because trmm loads m*n matrix  B, stores m*n matrix C, and loads triangle matrix A. If  HIPBLAS_SIDE_LEFT == side the triangle matrix is m*m, if HIPBLAS_SIDE_LEFT != side the triangle matrix is n*n. We can approximate and say half the values in the triangle matrix need to be loaded. This same approximation is made in the legacy blas timing code to calculate flops.

clients/include/testing_trmm_batched.hpp
```diff
@@ -163,6 +148,66 @@ hipblasStatus_t testing_trmm_batched(const Arguments& argus)
...
+                      e_N,
+                      e_lda,
+                      e_ldb,
+                      e_batch_count>{}
+            .log_args<T>(std::cout,
+                         argus,
+                         gpu_time_used,
+                         trmm_gflop_count<T>(M, N, K),
+                         trmm_gbyte_count<T>(M, N, K),
+                         hipblas_error);
```
amcamd, line 210: see comment above, should be

    trmm_gflop_count<T>(M,N,side)
    trmm_gbyte_count<T>(M,N,side)

clients/include/testing_trmm_strided_batched.hpp
```diff
@@ -136,6 +132,69 @@ hipblasStatus_t testing_trmm_strided_batched(const Arguments& argus)
...
+                      e_lda,
+                      e_ldb,
+                      e_stride_b,
+                      e_batch_count>{}
+            .log_args<T>(std::cout,
+                         argus,
+                         gpu_time_used,
+                         trmm_gflop_count<T>(M, N, K),
+                         trmm_gbyte_count<T>(M, N, K),
+                         hipblas_error);
```
amcamd, line 197: see comment above, should be

    trmm_gflop_count<T>(M,N,side)
    trmm_gbyte_count<T>(M,N,side)

clients/include/flops.hpp
daineAMD: This is almost the same as what I have currently. In `testign_trmm.hpp` we have `int K = (side == HIPBLAS_SIDE_LEFT ? M : N);`, so the only difference would be getting rid of the `+ 1`, which I can do.

clients/include/flops.hpp
daineAMD: Same here, I can get rid of the `+1` but I'm not sure I see any reason to change it to passing in `side` rather than `k`.

clients/include/bytes.hpp
daineAMD: I'm not including the `m * n` store to C right now, I'll add that. The numeber of elements in the triangular matrix is currently using our function `tri_count(k)` which is `(k * (k + 1)) / 2`. I find it odd that the flop calculation is different in the legacy blas, maybe I'm overlooking something. I guess it should be consistent between flops.hpp and bytes.hpp either way.

clients/include/flops.hpp
amcamd: In the flop count from legacy BLAS they keep the higher order term, so m * n * (n+1) = m * n * n + m * n = O(m * n * n) . The calculation of flop (or NOP) is correct in the limit of large size. I find it easiest to keep with the legacy BLAS calculation, if there is any question on the flop count, we can say we got it from the legacy blas code (and it has not been changed since it was first used in 1-July-1988).

PR #312, opened 2021-03-05
trmm legacy flop calculations

daineAMD:
Matching trmm flop calculations from http://www.netlib.org/blas/sblas3time.f. See #311 for some discussion.


amcamd (approved):
LGTM

daineAMD:
@amcamd FYI it appears that my latest commit (turning the timing feature off for trmm tests, unless set by hipblas-bench) fixes the issue in CI. Looks like it was stuck in a while loop since the `iters` params weren't correctly set since we aren't supposed to hit the timing section in hipblas-test.

PR #317, opened 2021-03-11
Adding iamax, iamin, and nrm2 to hipblas-bench

daineAMD:
I'll add all rot* functions to hipblas-bench in another PR, then update all other L1 functions to use hipblasLocalHandle in another to keep PRs short. After that all L1 functions should be good to go and I'll move forward to L2.


TorreZuk (approved):
LGTM if tests pass, just the one comment on if you are just testing the test in those early returns.

clients/include/testing_iamax_iamin_batched.hpp
```diff
@@ -23,137 +23,125 @@ hipblasStatus_t testing_iamax_iamin_batched(const Arguments&                 arg
...
+        device_batch_vector<T> dx(1, 1, 5);
+        device_vector<int>     d_hipblas_result(1);
 
-        status_1 = func(handle, N, dx_array, incx, batch_count, d_rocblas_result);
+        return func(handle, N, dx, incx, batch_count, d_hipblas_result);
     }
     else if(batch_count < 0)
     {
-        status_1 = HIPBLAS_STATUS_INVALID_VALUE;
+        return HIPBLAS_STATUS_INVALID_VALUE;
```
TorreZuk, line 40: Sorry I think Edge ate my comment here.   What is being tested here by the return HIPBLAS_STATUS_INVALID_VALUE, now there is a macro EXPECT_HIPBLAS_STATUS so maybe that could be used to call func and expect HIPBLAS_STATUS_INVALID_VALUE but then return SUCCESS?

daineAMD: Basically this line is useless right now for hipblas-test, as you said it's just testing the test itself. I don't think we're interested in testing edge cases in hipBLAS, but this line is at least somewhat-useful in hipblas-bench.

If we decide we want to test some edge cases in hipBLAS, I think it's a good idea to use the new macro. There is some logic currently in blas1_gtest.cpp which tests the return value, but it may be a better idea to use the macro here. One issue we would have to deal with is hipBLAS/cuBLAS support, for the batched functions we'll get a `NOT_IMPLEMENTED` value, while getting a `INVALID_VALUE` return from rocBLAS.

Anyway, I'll leave it for now as it's somewhat useful for hipBLAS bench to have the return here. This pattern is seen throughout the tests so it can be a library-wide change if we decide.

PR #323, opened 2021-03-24
Fix for nrm2_batched test

daineAMD:
Removed hipblasDestroy() calls, but didn't update to new handle, causing a memory leak. This should fix that.


TorreZuk (approved):
LGTM

PR #325, opened 2021-03-25
Removing unsupported tests on cuBLAS

daineAMD:
A side effect of adding the CHECK_HIPBLAS_ERROR macro was introducing failures with the cuBLAS backend. To circumvent this, I've `ifndef`'d out tests (mainly batched/strided_batched) tests that aren't relevant to cuBLAS as they return not_supported. Feel free to discuss if anyone has a better solution.

Also, after adding host/device tests in some cases, we've uncovered a discrepency between rocBLAS and hipBLAS:
- in her and her2 with alpha == 0 and alpha on the device, rocBLAS appears to return before making any changes (see [here](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/blob/develop/library/src/blas2/rocblas_her.hpp#L46)). This matches the result of the reference code. cuBLAS, however, still sets the imaginary portion of the diagonal to 0.
- The netlib documentation [here](http://www.netlib.org/lapack/explore-html/d6/dee/group__complex__blas__level2_gafa73370c613ec8f157771992010809ac.html#gafa73370c613ec8f157771992010809ac) mentions that the imaginary is set to 0 (matching cuBLAS' device implementation), but in the reference code, there is a quick return for alpha == 0 (matching rocBLAS and cuBLAS' host implementation).

To me, this seems to be a bug in cuBLAS since their result is dependent on whether or not alpha is passed in on the host or device, but we can discuss.


clients/gtest/axpy_ex_gtest.cpp
```diff
@@ -157,7 +161,7 @@ TEST_P(axpy_ex_gtest, axpy_batched_ex)
         }
         else
         {
-            EXPECT_EQ(HIPBLAS_STATUS_NOT_SUPPORTED, status); // for CUDA
+            EXPECT_EQ(HIPBLAS_STATUS_SUCCESS, status); // fail
```
TorreZuk, line 164: Would this not pass , what does // fail mean?
EXPECT_EQ( defined(__HIP_PLATFORM_NVCC__) ? HIPBLAS_STATUS_NOT_SUPPORTED :  HIPBLAS_STATUS_SUCCESS, status );

clients/include/testing_her2.hpp
```diff
@@ -95,7 +95,11 @@ hipblasStatus_t testing_her2(const Arguments& argus)
         if(argus.unit_check)
         {
             unit_check_general<T>(N, N, lda, hA_cpu.data(), hA_host.data());
-            unit_check_general<T>(N, N, lda, hA_cpu.data(), hA_device.data());
+
+            // NOTE: on cuBLAS, with alpha == 0 and alpha on the device, there is not a quick-return,
+            // instead, the imaginary part of the diagonal elements are set to 0. in rocBLAS, we are quick-returning
+            // as well as in our reference code. For this reason, I've disabled the check here.
+            // unit_check_general<T>(N, N, lda, hA_cpu.data(), hA_device.data());
```
TorreZuk, line 102: You could just guard with if (alpha) as you know what the device matches the host here.

clients/include/testing_her.hpp
```diff
@@ -91,7 +91,11 @@ hipblasStatus_t testing_her(const Arguments& argus)
         if(argus.unit_check)
         {
             unit_check_general<T>(N, N, lda, hA_cpu.data(), hA_host.data());
-            unit_check_general<T>(N, N, lda, hA_cpu.data(), hA_device.data());
+
+            // NOTE: on cuBLAS, with alpha == 0 and alpha on the device, there is not a quick-return,
+            // instead, the imaginary part of the diagonal elements are set to 0. in rocBLAS, we are quick-returning
+            // as well as in our reference code. For this reason, I've disabled the check here.
+            // unit_check_general<T>(N, N, lda, hA_cpu.data(), hA_device.data());
```
TorreZuk, line 98: if (alpha) ?

TorreZuk (commented):
Some ideas.  As for alpha==0  I vote for keeping device and host behaviour the same...

clients/gtest/axpy_ex_gtest.cpp
daineAMD: If I removed the `ifndef`, the code you suggested wouldn't work by itself. In testing_xx.hpp, when it reaches the first `CHECK_HIPBLAS_ERROR()` from a `_batched` call, there will be an error as on NVCC we will not be returning success. We could potentially change the macro to pass on NVCC if we get a `HIPBLAS_STATUS_SUCCESS` OR `HIPBLAS_STATUS_NOT_SUPPORTED`, but that could potentially hide issues as well.

The `// fail` comment means that if we get to that point in the code, the test will be failing. The `if` statement a couple lines prior avoids this code if we returned a success, this could be changed a bit/made more clear, but I won't do that here.

TorreZuk (approved):
Sure that sounds reasonable, but you can easily make another macro for the math function call only HIP_NO_ERROR and allow unsupported.  Then you still verify unsupported in the outside class which may require more changes.   Maybe a skip would get logged?

daineAMD:
I don't mind that, what do you think of something like this:

```
#ifdef __HIP_PLATFORM_NVCC__

#ifndef CHECK_HIPBLAS_ERROR
#define EXPECT_HIPBLAS_STATUS(status, expected, unsupported) \
    do                                                       \
    {                                                        \
        hipblasStatus_t status__ = (status);                 \
        if(status__ != expected)                             \
        {                                                    \
            // Added this if statement                       \
            if(status__ != unsupported)                      \
            {                                                \
                fprintf(stderr,                              \
                        "hipBLAS error: %s at %s:%d\n",      \
                        hipblasStatusToString(status__),     \
                        __FILE__,                            \
                        __LINE__);                           \
            }                                                \
            return (status__);                               \
        }                                                    \
    } while(0)
#define CHECK_HIPBLAS_ERROR(STATUS) EXPECT_HIPBLAS_STATUS(STATUS, HIPBLAS_STATUS_SUCCESS, HIPBLAS_STATUS_NOT_SUPPORTED)
#endif

#else // NOT __HIP_PLATFORM_NVCC__

// This doesn't change
#ifndef CHECK_HIPBLAS_ERROR
#define EXPECT_HIPBLAS_STATUS(status, expected)      \
    do                                               \
    {                                                \
        hipblasStatus_t status__ = (status);         \
        if(status__ != expected)                     \
        {                                            \
            fprintf(stderr,                          \
                    "hipBLAS error: %s at %s:%d\n",  \
                    hipblasStatusToString(status__), \
                    __FILE__,                        \
                    __LINE__);                       \
            return (status__);                       \
        }                                            \
    } while(0)
#define CHECK_HIPBLAS_ERROR(STATUS) EXPECT_HIPBLAS_STATUS(STATUS, HIPBLAS_STATUS_SUCCESS)
#endif

#endif
```

Then I would revert the changes to leave the check for unsupported in the outside class as you mentioned.

Alternatively, we could use the first macro in this comment for all platforms, the downside is that we would miss any unexpected `HIPBLAS_STATUS_NOT_SUPPORTED` returns with rocBLAS, unless we combined it with a platform specific check in the outside class.

TorreZuk:
That looks okay but It's up to you, I just think if we can verify unsupported cases for both implementations that makes sense to do it, does your change show up as PASSED correct?   That is fine by me too

daineAMD:
I've added a comment in [LWPMLSE-122](http://ontrack-internal.amd.com/browse/LWPMLSE-122). I'll merge this one for now since everything is passing here and develop/staging are currently broken. Can discuss again later.

PR #329, opened 2021-04-08
Improving exception handling

daineAMD:
Adding `catch`s to hipBLAS as discussed.
- introduces `HIPBLAS_STATUS_INVALID_ENUM`. If we don't want this, we can expand the definition of `HIPBLAS_STATUS_INVALID_VALUE`, but I prefer adding this additional status
- Instead of throwing strings in failed conversions, we now throw this status.
- Adding `exception_to_hipblas_status` (same as in rocBLAS)
- Added `try` `catch` blocks to hcc gbmv. Tested this out in hipblas-bench with a non-existent transpose type, seemed to work as expected.

The `try` `catch` blocks will have to be added to all hcc and nvcc functions. I'll see if I can do a simple find & replace to avoid some tedious work :)

### DIFF
--- a/library/include/exceptions.hpp
+++ b/library/include/exceptions.hpp
@@ -6,7 +6,7 @@
 
 // Convert the current C++ exception to hiblasStatus_t
 // This allows extern "C" functions to return this function in a catch(...) block
-// while converting all C++ exceptions to an equivalent rocblas_status here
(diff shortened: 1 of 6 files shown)


TorreZuk (approved):
LGTM but if search and replace fails feel free to split up the work.

tfalders (approved):
Awesome, thanks Daine!

PR #330, opened 2021-04-08
Add try and catch blocks to all functions

daineAMD:
Did a find and replace to add try{} and catch{} blocks for every function in hipblas.cpp (hcc and nvcc). Sorry for the large PR, but it seems to have worked after some cleanup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 ### Added
 - Added hipblasStatusToString
 
+### Fixed
+- Added catch() blocks around API calls to prevent the leak of C++ exceptions
+
 ## [hipBLAS 0.44.0 for ROCm 4.2.0]
 ### Added
 - Made necessary changes to work with rocBLAS' gemm_ex changes. When using rocBLAS backend, hipBLAS will query the preferable
--- a/clients/gtest/auxiliary_gtest.cpp
+++ b/clients/gtest/auxiliary_gtest.cpp
@@ -3,6 +3,7 @@
  *
  * ************************************************************************ */
 
+#include "testing_exceptions.hpp"
 #include "utility.h"
 #include <gtest/gtest.h>
 #include <math.h>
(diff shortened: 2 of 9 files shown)


TorreZuk (approved):
Excellent!  I assumed there were separate files for each function so it wouldn't have made any  sense to split up the work.

amcamd (approved):
LGTM, It is good to see the code quality incrementally improving by improvements like this preventing the leak of C++ exceptions from a C API.

TorreZuk:
Probably worth a mention in the changelog as well.

YvanMokwinski:
I was talking with Torre on this topic. This is something that needs to be standardized in {hip/roc}SPARSE too.

cgmb:
I assume that rocBLAS may throw from the `*_template` methods? We adopted this try/catch in a few places in rocSOLVER where the possibility of throwing was obvious, but I suppose it should probably be done across-the-board. If there's anything we _know_ can't throw, we can mark it `noexcept`.

YvanMokwinski:
May throw if using some class constructors ?  
In the rocSPARSE library the only thing that can throw is the C++ implementation of the handle (constructor), probably the same in rocBLAS? 

I remember Lee saying 'noexcept' is obsolete. I haven't verified.

cgmb:
As far as I know, `noexcept` is not obsolete. However, it doesn't help in ensuring program correctness. It's a promise from the programmer to the compiler (and to other programmers) that a function will not throw. That sort of guarantee can sometimes be useful for optimization, but breaking that promise results in a call to `std::terminate`, so marking a function as `noexcept` requires great care.

TorreZuk:
Daine is one of these failures that throws tested by the yaml or other unit test?   Sorry didn't have time to look for myself.

daineAMD:
> Daine is one of these failures that throws tested by the yaml or other unit test? Sorry didn't have time to look for myself.

I didn't add any unit tests, I just did a manual test by removing an enum value from a conversion and trying to pass it in via hipblas-bench. I could add tests if desired.

TorreZuk:
Could be a single call with a bad enum to convert that triggers the throw and catch.   Wouldn't say you have to test them all but verify that exception handling occurs as expected.  Or just sneak another case into a bad_args test that already exists.   Separate PR would be fine I would think.   If there isn't something like general_gtest.cpp from rocBLAS probably that might be worth adding for misc. tests.

daineAMD:
I've added a quick test that calls gemv with a bad hipblasOperation_t and ensures that the new error code is returned, let me know what you think. I can add tests for all conversions later.

TorreZuk:
Yes that looks good to me.     If something goes awry with exception handling that should catch it  :-)

PR #334, opened 2021-04-19
Removing gemm int8 tests

daineAMD:


TorreZuk (approved):
LGTM but either TODO or comment to explain as you may forget why this was done.

amcamd (approved):

PR #337, opened 2021-04-21
Removed getri_outofplace_batched declaration

tfalders:
getri_outofplace_batched will be added to rocSOLVER's public API in 4.3 by https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/243. This PR removes the declarations that were needed because of it not being part of the public API.


daineAMD (approved):
LGTM once CI runs with the updated rocSOLVER version and passes.

PR #343, opened 2021-05-12
Symv, tbmv, tbsv, tpmv, tpsv, trmv test updates

daineAMD:
Updating tests for symv, tbmv, tbsv, tpmv, tpsv, trmv. All of L2 should be in hipblas-bench now.


TorreZuk (approved):
Very good additions.  Didn't spot anything.

PR #345, opened 2021-06-04
enable build with code coverage

YvanMokwinski:


CMakeLists.txt
```diff
@@ -93,6 +93,14 @@ if( USE_CUDA )
     list( APPEND HIP_INCLUDE_DIRS "${HIP_ROOT_DIR}/include" )
 endif( )
 
+
+option(BUILD_CODE_COVERAGE "Build with code coverage enabled" OFF)
+if(BUILD_CODE_COVERAGE)
+  add_compile_options(-fprofile-arcs -ftest-coverage)
```
daineAMD, line 99: In rocBLAS it looks like we're defaulting to use debug mode if this option is passed in, probably want the same here.

CMakeLists.txt
```diff
@@ -104,3 +112,66 @@ set(CMAKE_INSTALL_LIBDIR "lib" CACHE INTERNAL "Installation directory for librar
...
+  # > make coverage GTEST_FILTER=<>
+  # will run:
+  #  > make coverage_analysis GTEST_FILTER=<> (analyze tests)
+  #  > make coverage_output (generate html documentation)
+  #
+
+  #
+  # Run coverage analysis
+  #
+  set(coverage_test ./clients/staging/hipblas-test)
```
daineAMD, line 132: Since we're only allowing this in debug mode, is this needed or can we just always use hipblas-test-d?

daineAMD (approved):
Sorry for the (very) late review. Checked it out and seems to work. Will definitely be more useful in rocBLAS, but still a good addition here too!

PR #351, opened 2021-06-18
hipblas on Windows

daineAMD:
See rocBLAS changes https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/661 for reference.


clients/common/utility.cpp
```diff
@@ -60,6 +78,28 @@ int type2int<hipblasDoubleComplex>(hipblasDoubleComplex val)
...
+        {
+            result.resize(length);
+            result.shrink_to_fit();
+            break;
+        }
+        result.resize(result.size() * 2);
+    }
+
+    std::wstring wspath(result.data());
+    std::filesystem::path exepath(wspath.begin(), wspath.end());
```
TorreZuk, line 97: This section should be changed to use the non UNICODE GetModuleFileNameA call for now, unless you decide to define UNICODE.   See other more recent changes in rocblas.  Can discuss but probably can get windows CI running as as long as you don't break linux this can be determined later.

clients/common/utility.cpp
```diff
@@ -60,6 +78,28 @@ int type2int<hipblasDoubleComplex>(hipblasDoubleComplex val)
...
+        if(length < result.size() - 1)
+        {
+            result.resize(length);
+            result.shrink_to_fit();
+            break;
+        }
+        result.resize(result.size() * 2);
+    }
+
+    std::wstring wspath(result.data());
```
cgmb, line 96: The `std::wstring` constructor that takes a pointer and a size is typically more efficient than the one that depends on null termination. That said, I think it would be even better if you could replace your `result` variable with `wspath` to eliminate this copy entirely.

(There also seems to be a bug here. The value returned by `GetModuleFileNameW` is the length excluding the null terminator, so the `result.resize(length);` earlier in the function must drop the null terminator. The `wspath` constructor will therefore read past the end of `result`.)

clients/common/utility.cpp
```diff
@@ -79,22 +79,22 @@ int type2int<hipblasDoubleComplex>(hipblasDoubleComplex val)
...
+    std::vector<TCHAR> result(MAX_PATH + 1);
     // Ensure result is large enough to accomodate the path
     for(;;)
     {
-        auto length = GetModuleFileNameW(nullptr, result.data(), result.size());
+        auto length = GetModuleFileNameA(nullptr, result.data(), result.size());
         if(length < result.size() - 1)
         {
             result.resize(length);
-            result.shrink_to_fit();
```
TorreZuk: You'll want the resize length + 1 as Cory noted it is actually not counting the null character for this success return value.

TorreZuk (approved):
I'll change rocBLAS to avoid the read just beyond the result as well. 
The other suggestion Cory made is just pass the pointer, length+1 to the string/path constructor is fine too.     If this is ready to build you can make the call to get win CI building as he can just pass in the noSOLVER flag right?

PR #358, opened 2021-07-26
static blis usage

daineAMD:
See https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/800


TorreZuk (approved):
You'll have to ask @DanyiLin to update the pipeline to allow BLIS_DIR to be set and point at the new directory.   Unfortunately this one seems slower but for now want 100% tests passing.

PR #359, opened 2021-07-26
Updating changelog

daineAMD:
Not many changes in hipBLAS this cycle, but thought there should be something there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log for hipBLAS
 
+## (Unreleased) hipBLAS 0.48.0
+### Added
+- Added more support for hipblas-bench
+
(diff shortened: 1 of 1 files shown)


TorreZuk (approved):
You could add if you want more?
#Fixed 
Avoid large offset overflow for gemv and hemv in hiblas-test

amcamd:
Could you also add:

Added hipblas-bench with support for syrkx

daineAMD:
@amcamd That's what I was trying to imply with "Added more support for hipblas-bench". There were a handful of functions added, I can list them all separately if desired, although I haven't done that in the past releases.

amcamd:
@daineAMD, sorry I should have seen your comment covers syrkx changes. Go ahead with your present PR, no need to list functions separately.

amcamd (approved):
LGTM

PR #363, opened 2021-07-29
Windows install and debug fixes

daineAMD:
Fixing debug and install flag for windows build.

I notice that the change I made in library/src/CMakeLists.txt for the debug build isn't the same as rocBLAS/rocSOLVER. I'm guessing this is because of [this line](https://github.com/ROCmSoftwarePlatform/hipBLAS/blob/develop/library/src/CMakeLists.txt#L112). Since I'm still getting the hang of cmake I thought I'd bring this up in case there's a better solution. The other changes match rocBLAS.


TorreZuk (approved):
We don't test debug in CI and as these build types go into separate build/debug or build/release the two variants won't co-exist.   If you want to remove set_target_properties( hipblas PROPERTIES DEBUG_POSTFIX "-d" ) then you wouldn't need the different name for the .pdb copy (pdb only for debug info).   Fine for me with this fix or to remove the -d suffix.  If we want to package release and debug together we may want the suffix down the road.   No advise on packaging yet, but I am assuming only release builds packaged to start, same as linux.

PR #368, opened 2021-08-24
Fix HIP_PATH check

daineAMD:
Accidentally closed https://github.com/ROCmSoftwarePlatform/hipBLAS/pull/365

See https://github.com/ROCmSoftwarePlatform/hipSOLVER/pull/41

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log for hipBLAS
 
+## (Unreleased) hipBLAS 0.49.0
+### Fixed
+- Fixed use of incorrect 'HIP_PATH' when building from source.
+
 ## (Unreleased) hipBLAS 0.48.0
 ### Added
 - Added more support for hipblas-bench
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,7 +63,7 @@ include( ROCMInstallSymlinks )
 set ( VERSION_STRING "0.49.0" )
 rocm_setup_version( VERSION ${VERSION_STRING} )
 
-if( NOT DEFINED $ENV{HIP_PATH})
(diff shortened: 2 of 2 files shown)


TorreZuk (approved):
I think that is correct syntax now but assume you tested.  You can swap the if/else code then you can drop the NOT and may read better.   
Now for windows support you may want to add env path convention converter helper:
file(TO_CMAKE_PATH "$ENV{HIP_PATH}" HIP_DIR)
or 
file(TO_CMAKE_PATH "$ENV{HIP_DIR}" HIP_DIR)
And as above consider change to use HIP_DIR local variable instead of HIP_PATH as HIP_PATH is something that hip cmake code sometimes sets.    Who is advising on setting the environment variable HIP_PATH I guess is the question, if it is HIP people then that is fine.  
On windows I think we are now promoting HIP_DIR as an environment variable override but would have to confirm this with other libraries.
The cmake detection of module vs. config (two different usage patters) when finding of hip is the much discussed issue which I'm not sure if you are seeing any issue with here?  This is handled later by the USE_CUDA variable I think

cgmb (approved):
I came away from my debug session in https://github.com/ROCmSoftwarePlatform/hipSOLVER/issues/40 with many of the concerns that @TorreZuk  had, but I wouldn't worry about those broader issues right now. IMO, it's not worth improving the logic piecemeal in hip{BLAS,SOLVER}.

In the near future, I would like to:

1. Figure out The Right Way™ to find HIP for either AMD or NVIDIA.
2. Encapsulate that logic in a CMake file (e.g., a find module).
3. Use that file in hip{BLAS,SOLVER,FFT,SPARSE,RAND,CUB}.

We can do a very detailed code review with all the relevant experts and stakeholders on that one single way.

PR #370, opened 2021-08-30
Nrm2_ex windows failures

daineAMD:
Seeing if this will remove the failures from [SWDEV-300988](https://ontrack-internal.amd.com/browse/SWDEV-300988).

It appears that the cpu result on windows is off by 1 from the cpu result we are getting on linux for this specific half-precision test case. I believe this is due to the conversion from hipblasHalf -> float -> hipblasHalf during the cpu calculations. This differs from the rocBLAS implementation as rocBLAS uses _Float16, which apparently isn't defined for g++ (see [here](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/commit/5c144635b237cd0e0f96e2895a835317ce52e68c#diff-972107ba5a8a0ad89499e64ffa0d015ffc0d59e9379b265d947904b13f4181a2R49)).

If anyone has any ideas for a better solution, I'm all ears! I can work on getting _Float16 to work with hipcc in the meantime.

### DIFF
--- a/clients/gtest/gemm_ex_gtest.cpp
+++ b/clients/gtest/gemm_ex_gtest.cpp
@@ -472,59 +472,69 @@ TEST_P(parameterized_gemm_batched_ex, standard_strided_batched)
     }
 }
 
-class parameterized_chunk_gemm_ex : public ::TestWithParam<gemm_ex_tuple>
-{
-protected:
-    parameterized_chunk_gemm_ex() {}
-    virtual ~parameterized_chunk_gemm_ex() {}
-    virtual void SetUp() {}
-    virtual void TearDown() {}
-};
+// TODO: Disabling some gemm int8 tests as not supported by rocBLAS for all architectures
+// class parameterized_chunk_gemm_ex : public ::TestWithParam<gemm_ex_tuple>
+// {
+// protected:
+//     parameterized_chunk_gemm_ex() {}
+//     virtual ~parameterized_chunk_gemm_ex() {}
+//     virtual void SetUp() {}
+//     virtual void TearDown() {}
+// };
 
-TEST_P(parameterized_chunk_gemm_ex, float)
-{
-    // GetParam return a tuple. Tee setup routine unpack the tuple
-    // and initializes arg(Arguments) which will be passed to testing routine
-    // The Arguments data struture have physical meaning associated.
-    // while the tuple is non-intuitive.
+// TEST_P(parameterized_chunk_gemm_ex, float)
+// {
+//     // GetParam return a tuple. Tee setup routine unpack the tuple
+//     // and initializes arg(Arguments) which will be passed to testing routine
+//     // The Arguments data struture have physical meaning associated.
+//     // while the tuple is non-intuitive.
 
-    Arguments arg = setup_gemm_ex_arguments(GetParam());
+//     Arguments arg = setup_gemm_ex_arguments(GetParam());
 
-    hipblasStatus_t status = testing_gemm_ex(arg);
+//     hipblasStatus_t status = testing_gemm_ex(arg);
 
-    // if not success, then the input argument is problematic, so detect the error message
-    if(status != HIPBLAS_STATUS_SUCCESS)
-    {
-        if(arg.M < 0 || arg.N < 0 || arg.K < 0)
-        {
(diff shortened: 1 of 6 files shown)


TorreZuk:
Is the accumulation as float?   I guess you could look at tolerance vs. equality but why would batch count influence this agreement?   More random input so more chance to be unlucky?

amcamd:
I am having difficulties unzipping the file in  SWDEV-300988 . If it is converting from float to half it should be rounding. It is possible there is incorrect rounding, and this would allow the answer to be off up to 1 ULP (unit in last place). When you say one off, is it one unit in the last place? for ieee32 this would typically be off in the 7th decimal place?

daineAMD:
Accumulation is float, yes. I assumed that the large batch count just gave it more input which happened to fail, but that isn't the case. After looking into this some more, I believe I've found the root of the issue.

The failing test has the following half-precision output:
```
cpu_result[b] evaluates to 22962,
gpu_result[b] evaluates to 22963
```
The CPU implementation converts the input vector into floats, calls cblas_snrm2(), and converts the result back to a hipblasHalf, where we then compare it to rocBLAS' output.
The comparison we currently have sets the generous allowable error to `vector_length * epsilon * cpu_result`, in this case that would be `7111 * epsilon * 22962`. Our results would certainly fit into this threshold, except that we are using std::numeric_limits<hipblasHalf>, which is returning 0 as hipblasHalf is just a uint16_t. I've made a change to use the execution type as the epsilon value, I hope that makes sense.

An aside:
- It looks like the initialization for the batched code here isn't working as expected, the input for each batch is the same, so we aren't seeing the failures there. This should be fixed in my other open PR.

TorreZuk:
Okay will look at your batch init PR when it appears.  Yes numeric limits should be done on the execution type anyway but that the float16 being a uint16 should have been 1 for epsilon I would have thought ?   So am curious if this is something wrong, but if zero actually that is good as it should correctly show up as a failed test rather than an false positive.
Also we shouldn't be multiplying by the cpu_result, it should be some small factor as the size * epsilon is supposed to allow for round off variation with arbitrary order.   Can you check what rocblas does and maybe follow that.

daineAMD:
rocBLAS does similar, multiplying it by N and the cpu_result as well (see [here](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/blob/develop/clients/include/blas1/testing_nrm2.hpp#L119)).

As for the batched PR, it's just using the host_batch_vector initialization in #369

TorreZuk:
Okay if that is what is done in rocblas fine by me to keep it the same, I think that approximation of using the result isn't the best for growing sums as overly generous, as the result is not typical of the intermediate results but maybe good enough for now.  It is probably too much effort to model the process.

daineAMD:
I've commented out some unused gemm_ex test code here as well as gtest is giving an error for an unused test case.

TorreZuk:
Okay I see that in rocblas now when I run a subset of tests as well.  I think the gtest update is defaulting to this behaviour.
I will likely propose we just defeat this as we often run subsets on demand and don't want all these "errors" sound okay?
So it is up to us to ensure all the tests expected are run rather than GTest to tell us something wasn't run.

daineAMD:
Yep, that sounds fair to me!

TorreZuk (approved):
Not sure if you should make a ticket to restore the TODO tests, otherwise I'll let you know when I have a PR for rocblas and you can probably copy the same style.  I think I have it done in a branch somewhere.

TorreZuk:
Probably another ticket we should make, a helper for this should do 
typename std::enable_if<!std::numeric_limits<T>::is_integer, bool>::type withinTolerance(T gold, T value, size_t opCount, T toleranceFactor) ...
and maybe just use gold as the scalar as you do to find the least significant digit.  norm2 we can't use alternating sign to keep it from growing, but could use HPL init mode.  You want to make it or want me to?

daineAMD:
I'm not sure I completely understand what you're suggesting, this helper would be for norm checks as we talked about here (and not for near checks)?
As for the hpl init, that makes sense to me. I can probably add that to the catch-all for small hipBLAS test improvements [LWPMLSE-122](https://ontrack-internal.amd.com/browse/LWPMLSE-122) unless you think otherwise.

TorreZuk:
Can move the discussion to that ticket, it is just to avoid epsilon use on integer types in general

PR #372, opened 2021-09-02
Update solver tests

daineAMD:
Adding rocSOLVER functions to hipblas-bench and cleaning up rocSOLVER tests.

The gflop and gbyte calculations are certainly incorrect, so I would appreciate some help with them :smile:

### DIFF
--- a/clients/benchmarks/CMakeLists.txt
+++ b/clients/benchmarks/CMakeLists.txt
@@ -131,8 +131,6 @@ else( )
   target_link_libraries( hipblas-bench PRIVATE ${CUDA_LIBRARIES} Threads::Threads )
 endif( )
 
-target_compile_definitions( hipblas-bench PRIVATE __BUILD_WITH_SOLVER__ )
-
 set_target_properties( hipblas-bench PROPERTIES DEBUG_POSTFIX "-d" CXX_EXTENSIONS NO )
 set_target_properties( hipblas-bench PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/staging" )
 #add_dependencies( hipblas-bench hipblas-bench-common )
--- a/clients/benchmarks/CMakeLists.txt
+++ b/clients/benchmarks/CMakeLists.txt
@@ -131,6 +131,8 @@ else( )
   target_link_libraries( hipblas-bench PRIVATE ${CUDA_LIBRARIES} Threads::Threads )
 endif( )
 
+target_compile_definitions( hipblas-bench PRIVATE __BUILD_WITH_SOLVER__ )
+
 set_target_properties( hipblas-bench PROPERTIES DEBUG_POSTFIX "-d" CXX_EXTENSIONS NO )
 set_target_properties( hipblas-bench PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/staging" )
 #add_dependencies( hipblas-bench hipblas-bench-common )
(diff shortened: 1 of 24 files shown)


amcamd:
Operation counts for both BLAS and LAPACK are in Appendix C (page 118) of http://www.netlib.org/lapack/lawnspdf/lawn41.pdf . I think it should have what you need for gflop calculations.

I am looking for gbyte information.

amcamd:
The C and Z functions are for complex, and their operation count is 4 times the operation count for the real S and D functions. This is similar to rocBLAS. The lawn41.pdf document has counts for real functions.

For byte count, the complex functions are 2 times the count for real functions.

amcamd:
bytes.hpp looks correct for real versions of geqrf, getrf, getri, getrs. For complex, they would need to be multiplied by 2.

clients/include/testing_getrf_batched.hpp
```diff
@@ -81,56 +69,69 @@ hipblasStatus_t testing_getrf_batched(const Arguments& argus)
...
 
-            if(argus.unit_check)
-            {
-                U      eps       = std::numeric_limits<U>::epsilon();
-                double tolerance = eps * 2000;
+        hipblas_error = norm_check_general<T>('F', M, N, lda, hA, hA1, batch_count);
+        if(argus.unit_check)
+        {
+            U      eps       = std::numeric_limits<U>::epsilon();
+            double tolerance = eps * 2000;
```
TorreZuk, line 106: future work: we could create getrf_tolerance function much like getrf_gflops

clients/include/testing_getri_batched.hpp
```diff
@@ -41,40 +40,28 @@ hipblasStatus_t testing_getri_batched(const Arguments& argus)
...
+    device_vector<int>     dIpiv(Ipiv_size);
+    device_vector<int>     dInfo(batch_count);
+
+    double             gpu_time_used, hipblas_error;
+    hipblasLocalHandle handle(argus);
 
     // Initial hA on CPU
+    hipblas_init(hA, true);
+
     srand(1);
```
TorreZuk, line 62: not sure who this srand is for?

clients/benchmarks/CMakeLists.txt
```diff
@@ -131,6 +131,8 @@ else( )
   target_link_libraries( hipblas-bench PRIVATE ${CUDA_LIBRARIES} Threads::Threads )
 endif( )
 
+target_compile_definitions( hipblas-bench PRIVATE __BUILD_WITH_SOLVER__ )
+
```
TorreZuk, line 135: underscores often used for special meaning, I would suggest just HIPBLAS_WITH_SOLVER ?

TorreZuk (approved):
LGTM, I would suggest waiting on a solver reviewer as I wasn't sure of their functions.  Another good cleanup too!

clients/benchmarks/CMakeLists.txt
TorreZuk: @daineAMD HIP by itself would indicate to me the hip layer so that is why I suggested simply "HIPBLAS_WITH_SOLVER"
This is some advice I found in a search "According to the C++ Standard, identifiers starting with one underscore are reserved for libraries. Identifiers starting with two underscores are reserved for compiler vendors."

daineAMD: I changed it to use __HIP_PLATFORM_SOLVER__ which was added in https://github.com/ROCmSoftwarePlatform/hipBLAS/commit/f9c8c54da7ef7bce2189f2bbb499138322ee3d4a. I can change it in another PR since it's probably out of the scope of this one.

PR #373, opened 2021-09-03
Moving test calls into if statement

daineAMD:
Moving hipBLAS test calls into `if(unit_check || norm_check)`. I don't see any use of making these calls unless we're testing the output, it should save some slight time when running benchmarks. This could be done to rocBLAS as well.

I've only done this for a few L1s to get feedback, if it's agreed upon I'll change all unit tests to use this. Also made some other small fixes along the way that I noticed.


TorreZuk (approved):
LGTM if tests pass, yes savings on the bench a good thing.

PR #376, opened 2021-09-10
Allowing negative increments in hipblas tests (L1)

daineAMD:
Allowing negative incx/incy inputs for testing (L1s for now). Also allows incx = 0 and incy = 0, although this typically gives incorrect results.

In rocBLAS we typically test that we still return success even if we're passing in nullptrs, I don't think that's in hipBLAS' scope to test that so I've opted not to add that for now.

- Asum - negative incx/incy not supported in rocBLAS, so not allowing testing here.
- Axpy - allowing negative incx/incy
- Copy - allowing negative incx/incy
- Dot - allowing negative incx/incy
- max/min - negative incx/incy not supported in rocBLAS, so not allowing testing here
- Nrm2 - negative incx/incy not supported in rocBLAS, so not allowing testing here
- Rot - allowing negative incx/incy
- Rotm - allowing negative incx/incy
- Scal - negative incx/incy not supported in rocBLAS, so not allowing testing here
- Swap - allowing negative incx/incy

### DIFF
--- a/clients/include/testing_axpy_batched_ex.hpp
+++ b/clients/include/testing_axpy_batched_ex.hpp
@@ -41,13 +41,13 @@ hipblasStatus_t testing_axpy_batched_ex_template(const Arguments& argus)
     Ta h_alpha = argus.get_alpha<Ta>();
 
     // Naming: dX is in GPU (device) memory. hK is in CPU (host) memory, plz follow this practice
-    host_batch_vector<Tx> hx(N, incx ? incx : 1, batch_count);
-    host_batch_vector<Ty> hy_host(N, incy ? incy : 1, batch_count);
-    host_batch_vector<Ty> hy_device(N, incy ? incy : 1, batch_count);
-    host_batch_vector<Ty> hy_cpu(N, incy ? incy : 1, batch_count);
+    host_batch_vector<Tx> hx(N, incx, batch_count);
+    host_batch_vector<Ty> hy_host(N, incy, batch_count);
+    host_batch_vector<Ty> hy_device(N, incy, batch_count);
+    host_batch_vector<Ty> hy_cpu(N, incy, batch_count);
 
-    device_batch_vector<Tx> dx(N, incx ? incx : 1, batch_count);
-    device_batch_vector<Ty> dy(N, incy ? incy : 1, batch_count);
+    device_batch_vector<Tx> dx(N, incx, batch_count);
+    device_batch_vector<Ty> dy(N, incy, batch_count);
     device_vector<Ta>       d_alpha(1);
 
     CHECK_HIP_ERROR(dx.memcheck());
--- a/clients/include/testing_axpy_batched_ex.hpp
+++ b/clients/include/testing_axpy_batched_ex.hpp
@@ -24,17 +24,31 @@ hipblasStatus_t testing_axpy_batched_ex_template(const Arguments& argus)
     int incy        = argus.incy;
     int batch_count = argus.batch_count;
 
-    // argument sanity check, quick return if input parameters are invalid before allocating invalid
-    // memory
-    if(N <= 0 || batch_count <= 0)
-    {
-        return HIPBLAS_STATUS_SUCCESS;
-    }
-
     hipblasDatatype_t alphaType     = argus.a_type;
     hipblasDatatype_t xType         = argus.b_type;
     hipblasDatatype_t yType         = argus.c_type;
     hipblasDatatype_t executionType = argus.compute_type;
+
+    hipblasLocalHandle handle(argus);
+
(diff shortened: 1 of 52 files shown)


clients/include/testing_axpy_batched_ex.hpp
```diff
@@ -40,21 +36,18 @@ hipblasStatus_t testing_axpy_batched_ex_template(const Arguments& argus)
...
 
     // Naming: dX is in GPU (device) memory. hK is in CPU (host) memory, plz follow this practice
-    host_batch_vector<Tx> hx(N, incx, batch_count);
-    host_batch_vector<Ty> hy_host(N, incy, batch_count);
-    host_batch_vector<Ty> hy_device(N, incy, batch_count);
-    host_batch_vector<Ty> hy_cpu(N, incy, batch_count);
+    host_batch_vector<Tx> hx(N, incx ? incx : 1, batch_count);
+    host_batch_vector<Ty> hy_host(N, incy ? incy : 1, batch_count);
+    host_batch_vector<Ty> hy_device(N, incy ? incy : 1, batch_count);
+    host_batch_vector<Ty> hy_cpu(N, incy ? incy : 1, batch_count);
```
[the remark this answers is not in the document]
daineAMD: Good point, I'll do that.
We don't mention anything in our API about inc == 0 for functions that support inc < 0. If it's writing to the vector with the 0 increment I would imagine it would be a race condition. If it's just reading it would just read the first element over and over - I don't see anything mentioned in the lapack [saxpy documentation](http://www.netlib.org/lapack/explore-html/df/d28/group__single__blas__level1_gad2a52de0e32a6fc111931ece9b39726c.html#gad2a52de0e32a6fc111931ece9b39726c)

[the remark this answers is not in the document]
TorreZuk: Well if you want to match cublas then you'll need to pass the 0 cases through and not just return success.  The allocs can also just allocate 1 entry if increment is 0.   Fine by me.  Maybe Andrew has an opinion on this?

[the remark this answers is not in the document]
daineAMD: Made a change for this, I haven't checked if it all passes on cuBLAS yet.

clients/include/testing_iamax_iamin_strided_batched.hpp
```diff
@@ -52,8 +52,8 @@ hipblasStatus_t testing_iamax_iamin_strided_batched(const Arguments&
 
         if(batch_count > 0)
         {
-            host_vector<int> cpu_0(1);
-            host_vector<int> gpu_0(1);
+            host_vector<int> cpu_0(batch_count);
+            host_vector<int> gpu_0(batch_count);
             CHECK_HIP_ERROR(hipMemcpy(
                 gpu_0, d_hipblas_result_0, sizeof(int) * batch_count, hipMemcpyDeviceToHost));
             unit_check_general<int>(1, batch_count, 1, cpu_0, gpu_0);
```
TorreZuk, line 59: Can't select the line 62, but why is success returned and not the func status ?

daineAMD: We could, but it shouldn't make a difference at this point since we have the CHECK_HIPBLAS_ERROR. I think the status return in hipBLAS is kind of moot now since we have that macro, could change it to return void like rocBLAS at some point I think.

TorreZuk: Okay, so all the expected values are internal now.  I'll have to double check rocblas,  I guess the return should be a gtest expected value, not sure what they support again or how the test harness is used so will have to review that first. thanks

clients/include/testing_dot_strided_batched_ex.hpp
```diff
@@ -27,27 +27,63 @@ hipblasStatus_t testing_dot_strided_batched_ex_template(const Arguments& argus)
...
+                                                         d_hipblas_result_0,
+                                                         resultType,
+                                                         executionType));
+
+        if(batch_count > 0)
+        {
+            host_vector<Tr> cpu_0(batch_count);
+            host_vector<Tr> gpu_0(batch_count);
+            CHECK_HIP_ERROR(hipMemcpy(
+                gpu_0, d_hipblas_result_0, sizeof(Tr) * batch_count, hipMemcpyHostToDevice));
```
TorreZuk, line 81: DeviceToHost to get results!

TorreZuk (approved):
LGTM, but with the same flaw as rocblas I think, the comparisons using absolute value only work with negative one increments or certain combinations.  Maybe a comment in the gtest files to where {-1, -1} is set to say such and I think the comment above about incy > 0 is now wrong.

PR #380, opened 2021-09-14
Merging release/21.40 into develop after fixes

daineAMD:


TorreZuk (approved):
Only delta is packaging fix which is fine and windows only

PR #381, opened 2021-09-16
Allowing negative increments in hipblas tests (L2-1) 

daineAMD:
Allowing negative incx/incy and adding tests for bad inputs. Doing half of the L2s here, will do another PR for the rest.

@TorreZuk After thinking some more about [your comment](https://github.com/ROCmSoftwarePlatform/hipBLAS/pull/376#discussion_r708450335), I think I agree, we should be using the gtest macros. I've done that for these changes, let me know what you think.

I've made these changes here for: gbmv, gemv, hbmv, hemv, her, her2, hpmv, hpr, hpr2, trmv, tpmv, and tbmv.

### DIFF
--- a/clients/include/testing_gemv.hpp
+++ b/clients/include/testing_gemv.hpp
@@ -43,15 +43,24 @@ hipblasStatus_t testing_gemv(const Arguments& argus)
         dim_y = N;
     }
 
+    hipblasLocalHandle handle(argus);
+
     // argument sanity check, quick return if input parameters are invalid before allocating invalid
     // memory
-    if(M < 0 || N < 0 || lda < 0 || incx <= 0 || incy <= 0)
+    bool invalid_size = M < 0 || N < 0 || lda < M || lda < 1 || !incx || !incy;
+    if(invalid_size || !M || !N)
     {
-        return HIPBLAS_STATUS_INVALID_VALUE;
+        hipblasStatus_t actual = hipblasGemvFn(
+            handle, transA, M, N, nullptr, nullptr, lda, nullptr, incx, nullptr, nullptr, incy);
+        EXPECT_HIPBLAS_STATUS(
+            actual, (invalid_size ? HIPBLAS_STATUS_INVALID_VALUE : HIPBLAS_STATUS_SUCCESS));
+        return actual;
     }
 
-    X_size = dim_x * incx;
-    Y_size = dim_y * incy;
+    int abs_incx = incx >= 0 ? incx : -incx;
+    int abs_incy = incy >= 0 ? incy : -incy;
+    X_size       = dim_x * abs_incx;
+    Y_size       = dim_y * abs_incy;
 
     // Naming: dK is in GPU (device) memory. hK is in CPU (host) memory
     host_vector<T> hA(A_size);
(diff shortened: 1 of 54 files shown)


TorreZuk (commented):
Can let Andrew also look, that expect status look fine so you get it failing out there.
Returning the actual I guess could later be used to report a summary statistic total invalid, total succes, .... 
As for increment -2 do those functions all force to absolute value internally, otherwise how is the validation (norm/unit_check_general) work with abs_inc when the computation may be on different indices (0, 2), vs. (1, 3)?

clients/include/testing_gemv.hpp
```diff
@@ -43,15 +43,28 @@ hipblasStatus_t testing_gemv(const Arguments& argus)
...
-        return HIPBLAS_STATUS_INVALID_VALUE;
+        // cuBLAS doesn't seem to work with nullptrs in some cases here
+        /*
+        hipblasStatus_t actual = hipblasGemvFn(
+            handle, transA, M, N, nullptr, nullptr, lda, nullptr, incx, nullptr, nullptr, incy);
+        EXPECT_HIPBLAS_STATUS(
+            actual, (invalid_size ? HIPBLAS_STATUS_INVALID_VALUE : HIPBLAS_STATUS_SUCCESS));
+        return actual;
+	*/
+        return invalid_size ? HIPBLAS_STATUS_INVALID_VALUE : HIPBLAS_STATUS_SUCCESS;
```
amcamd, line 61: Does this pass with rocBLAS? I guess it can be tested in rocBLAS then the test can be left out of hipBLAS.

daineAMD: Yes, rocBLAS acts as I had expected, the commented code passes.

daineAMD:
An unrelated symm_strided_batched failure is seen in this CI run. I'm rerunning it to see if it shows up again, I've made a note of it and will look into it later this week.

clients/include/testing_gemv_batched.hpp
```diff
@@ -46,23 +46,32 @@ hipblasStatus_t testing_gemv_batched(const Arguments& argus)
...
+    hipblasLocalHandle handle(argus);
 
     // argument sanity check, quick return if input parameters are invalid before allocating invalid
     // memory
-    if(M < 0 || N < 0 || lda < 0 || incx <= 0 || incy <= 0 || batch_count < 0)
-    {
-        return HIPBLAS_STATUS_INVALID_VALUE;
-    }
-    else if(batch_count == 0)
+    bool invalid_size = M < 0 || N < 0 || lda < M || lda < 1 || !incx || !incy || batch_count < 0;
```
NaveenElumalaiAMD, line 53: Curious to know if  `lda < 1` is required?

daineAMD: This is directly from rocBLAS, so it appears that in the case that  `M == 0 && lda == 0`, we return `rocblas_status_invalid_size` as 0 is not an acceptable `lda`. However, if `M == 0 && lda > 0`, then we "quick return" `rocblas_status_success`.

clients/include/testing_gemv.hpp
amcamd: Can you replace "cuBLAS dosen't seem to work" with "Some platforms do not work".

TorreZuk: > Can you replace "cuBLAS dosen't seem to work" with "Some platforms do not work".

Might be clearer to say "Only rocblas conforms to expected behaviour so commenting out."

amcamd: Yes, Torre's wording is better, more descriptive and will be easier to understand.

PR #382, opened 2021-09-20
Modify hipBLAS level2 BLAS GFLOPs

NaveenElumalaiAMD:
This PR modifies the GFLOPs calculation of hipBLAS level 2 functions to be identical to level 2 rocBLAS functions.


daineAMD (commented):
I think ger, her, and her2 still need to be updated to match rocBLAS and blas2time. otherwise looks good.

daineAMD (approved):
LGTM as long as everything passes!

PR #385, opened 2021-09-21
Opt-in for using hipComplex types

daineAMD:
See https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/863. The sample code is a direct copy from there.

### DIFF
--- a/clients/samples/example_hip_complex_her2.cpp
+++ b/clients/samples/example_hip_complex_her2.cpp
@@ -100,13 +100,13 @@ int main(int argc, char** argv)
 
     for(int i1 = 0; i1 < N; i1++)
     {
-        hX[i1 * incx] = hipFloatComplex(1.0f, 0.0f);
-        hY[i1 * incy] = hipFloatComplex(1.0f, 0.0f);
+        hX[i1 * incx] = make_hipFloatComplex(1.0f, 0.0f);
+        hY[i1 * incy] = make_hipFloatComplex(1.0f, 0.0f);
     }
 
     for(int i1 = 0; i1 < rows; i1++)
         for(int i2 = 0; i2 < cols; i2++)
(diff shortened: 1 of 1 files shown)


library/include/hipblas.h
```diff
@@ -184,17 +154,67 @@ typedef struct hipblasDoubleComplex
...
-static_assert(std::is_trivial<hipblasComplex>{},
-              "hipblasComplex is not a trivial type, and thus is incompatible with C.");
-static_assert(std::is_trivial<hipblasDoubleComplex>{},
-              "hipblasDoubleComplex is not a trivial type, and thus is incompatible with C.");
+
+// Question: Do we need this check?
+// static_assert(std::is_trivial<hipblasComplex>{},
+//               "hipblasComplex is not a trivial type, and thus is incompatible with C.");
+// static_assert(std::is_trivial<hipblasDoubleComplex>{},
+//               "hipblasDoubleComplex is not a trivial type, and thus is incompatible with C.");
```
TorreZuk, line 216: Does it fail?   It must have passed before no?    The sizeof being 2 * floats etc.  may be equivalent ...

daineAMD: It fails with hipComplex types. I'm not sure why we need the check myself [cpp reference](https://en.cppreference.com/w/cpp/types/is_trivial)

TorreZuk: I think you don't want any default constructor doing anything different than C, for us we are covered by the C++ version  based definitions so maybe just move it into the #else our internal type checking.   Those who opt-into hip complex get what they asked for and the size compatibility is all we want to verify and standard layout.

TorreZuk (approved):
LGTM if all tests pass, but we need to wait for general approval.

evetsso (approved):
I have no objections.  rocFFT uses void*'s in its public API.  hipFFT already uses hipComplex and hipDoubleComplex, and does not expose any custom types.

PR #386, opened 2021-09-21
Adding compilation warning for trmm changes

daineAMD:
Not planning on merging this until later when we at least have a confirmed rocblas name. Let me know if the wording is OK or if anyone has any suggestions.

### DIFF
--- a/library/src/hcc_detail/hipblas.cpp
+++ b/library/src/hcc_detail/hipblas.cpp
@@ -13520,6 +13520,12 @@ catch(...)
 }
 
 // trmm
+// clang-format off
+#pragma GCC warning "The hipblasXtrmm API, along with batched versions, will \
+be changing in a future release to allow in-place and out-of-place behavior. This change \
+will introduce an output matrix \"C\", matching the rocblas_xtrmm_outofplace API and the \
+cublasXtrmm API."
+// clang-format on
 hipblasStatus_t hipblasStrmm(hipblasHandle_t    handle,
                              hipblasSideMode_t  side,
                              hipblasFillMode_t  uplo,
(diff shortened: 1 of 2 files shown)


TorreZuk (commented):
You should create an export helper with cmake with deprecation or manually make a define for HIPBLAS_DEPRECATED(MSG) __attribute__ ((deprecated(#MSG)))
A user may not want to see this more than once so can turn off with -wno-deprecated-declarations

daineAMD:
@TorreZuk I see that we have an export helper [here](https://github.com/ROCmSoftwarePlatform/hipBLAS/blob/develop/library/src/CMakeLists.txt#L124). This doesn't generate a deprecation define with a message, only `#  define HIPBLAS_DEPRECATED __attribute__ ((__deprecated__))`. I was wondering if you knew of a way to generate a define with a message, like `#  define HIPBLAS_DEPRECATED_MSG(MSG) __attribute__ ((deprecated(#MSG)))` as you suggested? I'm trying to read up on this but I thought you might be a quicker source.

TorreZuk:
Doesn't
# define HIPBLAS_DEPRECATED_MSG(MSG) __attribute__ ((deprecated(#MSG)))
work?

daineAMD:
Yep that works if I define it myself, was just wondering if there was a way to get it in the cmake generated file with the other definitions.

TorreZuk:
Well that you would have to look for as I don't know what is in rocm-cmake and what is in the cmake generator.  maybe no option for this message feature, but yours looks fine.

daineAMD:
No problem, just thought you might know. I'll just add the definition myself.

TorreZuk:
Still not sure of this.  It is not standard deprecation as you are not saying move off the API as there is no replacement to move to.   You are suggesting there will be a breaking change in the future, I guess someone may find it useful but not sure who.

amcamd (approved):
LGTM, I think this meets the requirement that we deprecate a function before changing it.

PR #394, opened 2021-10-13
Gemm bench to use host mode

daineAMD:
Changing gemm to use host mode when running benchmarks. This has better performance and matches rocBLAS.


TorreZuk (approved):
You should comment it has better perf. as for gemm we need to pull alpha/beta to host in device mode.

PR #401, opened 2021-10-19
L3 documentation

daineAMD:
Adding documentation for L3 functions. Find and replace worked well, need to go through everything to double check.

### DIFF
--- a/library/include/hipblas.h
+++ b/library/include/hipblas.h
@@ -5024,7 +5024,7 @@ HIPBLAS_EXPORT hipblasStatus_t hipblasCherkxStridedBatched(hipblasHandle_t      
 
     \details
 
-    herkxstridedBatched performs a batch of the matrix-matrix operations for a Hermitian rank-k update
+    herkxStridedBatched performs a batch of the matrix-matrix operations for a Hermitian rank-k update
 
     C_i := alpha*op( A_i )*op( B_i )^H + beta*C_i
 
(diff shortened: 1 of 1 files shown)


library/include/hipblas.h
```diff
@@ -4490,7 +4490,73 @@ HIPBLAS_EXPORT hipblasStatus_t hipblasZtrsvStridedBatched(hipblasHandle_t
...
 // ================================
 
-// herk
+/*! \brief BLAS Level 3 API
+
+    \details
+
+    herk performs one of the matrix-matrix operations for a Hermitian rank-k update
+
+    C := alpha*op( A )*op( A )^H + beta*C
```
TorreZuk, line 4499: Should follow the rocblas pattern, the doc comes after the type variants for some reason @amcamd .
  We also indicate multiple matches with a convention,    xHERK would be the name referred to above following rocblas.      We are lazy but the list might even be better as you google by function name so a perfect match is best for the user:   "herk (hipblasCherk and hipbasZherk) perform one of the ...  " we want a user search to go directly to our documentation on the web.
--  @rkamd

daineAMD: You're right, for rocBLAS I think the layout in readthedocs worked better if it was before the final variant. I'll make these changes.

amcamd: @TorreZuk  is correct, to produce the best output from parsing the doxygen comments they should go between the second last and last function prototype for that function. In this case, the doxygen comments would be between the hipblasCherk and hipblasZherk prototypes. You can see this in rocBLAS here: https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/blob/develop/library/include/internal/rocblas-functions.h#L9702 . Where a function has s, d, c, z datatypes,  the doxygen comments would be between the second last c and the last z prototypes. This applies also for the batched and strided batched functions.  @daineAMD @rkamd 

This will also apply to @rkamd 's PR #403. Thanks for catching this Torre, I missed this in review of #403

daineAMD: I changed the name to always just the blas name (herk, herkBatched, symm, symmStridedBatched, etc.). If anyone has strong feelings otherwise we can discuss.

PR #403, opened 2021-10-20
Documentation for BLAS Level-1 APIs

rkamd:
Added doxygen style documentation for level -1 API
converted function arguments to camel case

resolves #___

Summary of proposed changes:
-
-
-


TorreZuk:
I would think we can consider the function itself is singular so the "result" is valid for batched results as well.   Daine any preference before I start on L2 ?

daineAMD:
I don't really have a preference, either way is fine by me. I suppose being consistent with rocBLAS is ideal, though.

PR #405, opened 2021-10-21
Solver documentation

daineAMD:
Adding API documentation to solver functions and gemm since it was misplaced previously.

Github did a very bad job on the diff since I moved the gemm prototype unfortunately.

### DIFF
--- a/library/include/hipblas.h
+++ b/library/include/hipblas.h
@@ -13805,7 +13805,7 @@ HIPBLAS_EXPORT hipblasStatus_t hipblasCgetrfBatched(hipblasHandle_t       handle
     n-by-n matrices using partial pivoting with row interchanges. The LU factorization can
     be done without pivoting if ipiv is passed as a nullptr.
 
-    In the case that ipiv is null, the factorization of matrix \f$A_i\f$ in the batch has the form:
+    In the case that ipiv is not null, the factorization of matrix \f$A_i\f$ in the batch has the form:
 
     \f[
         A_i = P_iL_iU_i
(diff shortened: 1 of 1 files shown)


library/include/hipblas.h
```diff
@@ -13287,47 +13719,140 @@ HIPBLAS_EXPORT hipblasStatus_t hipblasDgetrf(
...
+    @param[out]
+    ipiv      pointer to int. Array on the GPU (the size depends on the value of strideP).\n
+              Contains the vectors of pivot indices ipiv_i (corresponding to A_i).
+              Dimension of ipiv_i is min(m,n).
+              Elements of ipiv_i are 1-based indices.
+              For each instance A_i in the batch and for 1 <= j <= min(m,n), the row j of the
+              matrix A_i was interchanged with row ipiv_i[j].
+              Matrix P_i of the factorization can be derived from ipiv_i.
+    @param[in]
+    strideP   hipblasStride.\n
```
rkamd, line 13837: Remove strideP. It is not in the list of function arguments

library/include/hipblas.h
```diff
@@ -13623,18 +14458,77 @@ HIPBLAS_EXPORT hipblasStatus_t hipblasCgeqrfBatched(hipblasHandle_t       handle
...
+    A         Array of pointers to type. Each pointer points to an array on the GPU of dimension lda*n.\n
+              On entry, the m-by-n matrices A_j to be factored.
+              On exit, the elements on and above the diagonal contain the
+              factor R_j. The elements below the diagonal are the last m - i elements
+              of Householder vector v_(j_i).
+    @param[in]
+    lda       int. lda >= m.\n
+              Specifies the leading dimension of matrices A_j.
+    @param[out]
+    ipiv      pointer to type. Array on the GPU.\n
```
tfalders, line 14513: Array of pointers to type. Each pointer points to an array on the GPU of dimension min(m, n).

tfalders (changes requested):
It looks like the solver documentation is mostly copied from rocSOLVER, but there are some important differences with the hipBLAS versions of these functions. These are the ones I caught.

PR #421, opened 2021-11-01
Removing uninstantiated gtest failures for cuda

daineAMD:


TorreZuk (approved):
Fine by me, so this matches the other functions?   Only non __HIP_PLATFORM_NVCC__

daineAMD:
It looks like most other functions don't have separate classes for batched/strided_batched versions, so we aren't getting the warning since nvcc always has the default non-batched one instantiated. I could change this to be consistent across the library but I don't think it's urgent at all.

PR #426, opened 2021-11-15
add rocsolver_path to install.sh, similar to rocblas_path

amcamd:
- add --rocsolver_path flag to install.sh.
- it is similar to --rocblas_path, and it allows for specifying a path to a custom rocSOLVER on your machine
- I have tested this on my machine and it worked


daineAMD (approved):

TorreZuk (approved):
LGTM

PR #427, opened 2021-11-17
remove cmake3 and cmake dependency from install.sh

amcamd:
- For  SWDEV-311824  cmake3.20.2 is being installed to replace the existing cmake. 
- This change should prevent this reinstall of a more recent version of cmake. This is similar to rocBLAS-internal PR 616
- This PR has only been tested manually on ubuntu. The CI will be used to test on other Linux distributions.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -125,11 +125,11 @@ install_packages( )
   fi
 
   # dependencies needed for library and clients to build
-  local library_dependencies_ubuntu=( "make" "cmake-curses-gui" "pkg-config" )
-  local library_dependencies_centos=( "epel-release" "make" "cmake3" "gcc-c++" "rpm-build" )
-  local library_dependencies_centos8=( "epel-release" "make" "cmake3" "gcc-c++" "rpm-build" )
-  local library_dependencies_fedora=( "make" "cmake" "gcc-c++" "libcxx-devel" "rpm-build" )
-  local library_dependencies_sles=( "make" "cmake" "gcc-c++" "libcxxtools9" "rpm-build" )
+  local library_dependencies_ubuntu=( "make" "pkg-config" )
+  local library_dependencies_centos=( "epel-release" "make" "gcc-c++" "rpm-build" )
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -423,12 +423,6 @@ fi
 # Default cmake executable is called cmake
 cmake_executable=cmake
 
-case "${ID}" in
-  centos|rhel)
-  cmake_executable=cmake3
```
TorreZuk, line 428: This is required on Centos as far as I know as most cmake installs are cmake3 not cmake.

TorreZuk (commented):
Andrew the CI does not test install -d you have to manually test yourself on all OS with dockers (yes a pain).   I would make sure we also add the option for cmake install as it is required above the default version for an OS so we should have the flag the same as rocblas.

install.sh
amcamd: This was removed in rocBLAS-internal PR 616, so rocBLAS no longer has this change to cmake3 for centos and rhel. I will test it on centos and rhel.

TorreZuk: @daineAMD I guess you did that, do you recall that cmake3 was retired?   Or I guess if we are not going with the default anymore the new install is just cmake and not the default cmake3 (symlink) ?

daineAMD: I unfortunately don't recall the details of removing cmake3, and don't know any details of cmake versioning. I think Alex ran into some issue on Centos with the oly cmake_executable=cmake3 in rocBLAS.

amcamd: there are come comments in https://ontrack-internal.amd.com/browse/SWDEV-311824   from Zhongyu saying cmake3 is an alias to cmake in Centos8

TorreZuk: Anyway, sounds fine to remove if done on rocblas they both either work or don't :-)   If you pull dockers you can see if Centos6 and 7 had cmake3 symlink to version 3.X and cmake was symlink to  2.X  ?   If user runs into trouble they can add a cmake3 symlink.

daineAMD (approved):

saadrahim (approved):

PR #434, opened 2021-11-24
hip to HIP for cuda install

daineAMD:


TorreZuk (approved):
LGTM

cgmb:
I was wondering if this was needed in hipSOLVER. Turns out the answer is: no. This is reverting a change made in #351, and hipSOLVER was forked from hipBLAS well before then.

PR #438, opened 2021-12-03
Doxygen grouping in hipBLAS

rkamd:
Add doxygen grouping to hipblas.h through the use of @{ and @} markers.

### DIFF
--- a/docs/source/functions.rst
+++ b/docs/source/functions.rst
@@ -108,870 +108,1310 @@ hipBLAS Functions
 Level 1 BLAS
 ============
 
+.. contents:: List of Level-1 BLAS Functions
+   :local:
+   :backlinks: top
+
 hipblasIXamax + Batched, StridedBatched
 -----------------------------------------
 .. doxygenfunction:: hipblasIsamax
+    :outline:
 .. doxygenfunction:: hipblasIdamax
+    :outline:
 .. doxygenfunction:: hipblasIcamax
+    :outline:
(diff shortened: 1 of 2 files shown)


docs/source/functions.rst
```diff
@@ -108,870 +108,1310 @@ hipBLAS Functions
...
 ============
 
+.. contents:: List of Level-1 BLAS Functions
+   :local:
+   :backlinks: top
+
 hipblasIXamax + Batched, StridedBatched
 -----------------------------------------
 .. doxygenfunction:: hipblasIsamax
+    :outline:
```
daineAMD, line 118: Do you have a link (or just an explanation) for this line? Tried some googling but couldn't find anything.

daineAMD (approved):

docs/source/functions.rst
rkamd: You can find more information [here](https://breathe.readthedocs.io/en/latest/directives.html). 

Without the `outline` directive, the generated PDF would repeat the description for each of the grouped functions. 
For example: 
```text
BLAS Level 1 API.
amax finds the first index of the element of maximum magnitude of a vector x.
.
.
```
Without the  `outline` directive the above code description will repeat for `hipblasIsamax`, `hipblasIdamax` , `hipblasIcamax` etc.

daineAMD: Understood, thanks!

PR #440, opened 2021-12-08
Updating prerequisites

daineAMD:


docs/source/install.rst
```diff
@@ -5,7 +5,15 @@ Building and Installing
 Prerequisites
 =============
 
--  A ROCm enabled platform, more information `here <https://rocm.github.io/>`_.
+* If using the rocBLAS backend on an AMD machine:
+
+  * A ROCm enabled platform, more information `here <https://rocm.github.io/>`_.
+  * A compatible version of rocBLAS
+
```
TorreZuk, line 12: Do you want to mention, optionally a compatible version of rocSOLVER for more functionality

TorreZuk (approved):

rkamd (approved):

docs/source/install.rst
```diff
@@ -5,7 +5,16 @@ Building and Installing
...
--  A ROCm enabled platform, more information `here <https://rocm.github.io/>`_.
+* If using the rocBLAS backend on an AMD machine:
+
+  * A ROCm enabled platform, more information `here <https://rocm.github.io/>`_.
+  * A compatible version of rocBLAS
+  * A compatible version of rocSOLVER for full functionality
+
+* If using the cuBLAS backend on a Nvidia machine:
+
+  * A HIP enabled platform, more information `here <https://rocmdocs.amd.com/en/latest/Installation_Guide/HIP-Installation.html#nvidia-platform/>`_.
```
cgmb, line 16: The slash that you often see at the end of a URL is a representation of the directory. For example, in the case of `https://rocm.github.io/`, that's the root directory.

This URL ends with a fragment-id (noted by the `#`). At that point, there shouldn't be any more slashes. The one on the end here actually breaks the reference to the `nvidia-platform` anchor (at least in Firefox).

docs/source/install.rst
```diff
@@ -5,7 +5,16 @@ Building and Installing
 Prerequisites
 =============
 
--  A ROCm enabled platform, more information `here <https://rocm.github.io/>`_.
+* If using the rocBLAS backend on an AMD machine:
+
+  * A ROCm enabled platform, more information `here <https://rocm.github.io/>`_.
```
cgmb, line 10: This URL just redirects to `https://rocmdocs.amd.com/en/latest/`.

cgmb (commented):
This is definitely an improvement, but one suggestion I might have for the future is to try to use descriptive link text for hyperlinks. The page or section title is usually a better choice than 'here'.

Good link text is particularly helpful for non-visual processing of web documents, notably search engines and screen readers. There's lots of good information out there if you search for "hyperlink guidelines" but an authoritative source would be the [Web Content Accessibility Guidelines § 2.4.4](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html), from the World Wide Web Consortium (W3C).

PR #444, opened 2022-01-14
changelog update for hipBLAS version 0.50

rkamd:
Changelog  update for hipBLAS version 0.50

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,6 @@
 - Added --cmake_install command line option to update cmake to minimum version if required
 - Added cmake-arg parameter to pass in cmake arguments while building
 - Added infrastructure to support readthedocs hipBLAS documentation.
-- Doxygen grouping of function in hipblas.h header file.
 
 ### Fixed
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -7,10 +7,13 @@
   absolute or relative path
 - Added --cmake_install command line option to update cmake to minimum version if required
 - Added cmake-arg parameter to pass in cmake arguments while building
+- Added infrastructure to support readthedocs hipBLAS documentation.
+- Doxygen grouping of function in hipblas.h header file.
```
TorreZuk, line 11: This isn't the clearest.   Wonder if ### Documentation section would be better

TorreZuk (commented):
LGTM

CHANGELOG.md
daineAMD: The [Changelog Document](https://confluence.amd.com/display/MLSE/Changelogs+and+Release+Notes) is probably a good reference, it doesn't mention a Documentation section, but I'm not sure if they're trying to enforce that layout or anything.

I think keeping the first bullet point and removing the second might be a better option.

PR #447, opened 2022-01-19
Batched gemmEx int8 fix

daineAMD:
Potential fix for SWDEV-315039.

It looks like the batched versions of gemm_ex int8 fail on some architectures as they do not query the gemm flag for rocBLAS. This PR hopefully fixes that, and adds back tests for int8. Currently aimed at a master hotfix as I didn't find the problem until after I finished feature complete, unfortunately. Will have to wait for CI to ensure this works on all architectures, including cuBLAS backend.


zaliu (approved):

amcamd (approved):

TorreZuk (approved):
LGTM if all tests pass

PR #450, opened 2022-01-25
User-specified initialization for hipblas-bench

daineAMD:
First commit  for adding NaN testing to hipBLAS. Based off the changes in rocBLAS.

### DIFF
--- a/clients/gtest/blas1_gtest.cpp
+++ b/clients/gtest/blas1_gtest.cpp
@@ -48,7 +48,6 @@ using ::testing::Combine;
 using ::testing::TestWithParam;
 using ::testing::Values;
 using ::testing::ValuesIn;
-using namespace std;
 
 // only GCC/VS 2010 comes with std::tr1::tuple, but it is unnecessary,  std::tuple is good enough;
 typedef std::tuple<int, vector<double>, vector<int>, double, int, bool> blas1_tuple;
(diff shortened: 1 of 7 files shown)


clients/gtest/blas1_gtest.cpp
```diff
@@ -2404,6 +2409,56 @@ TEST_P(blas1_gtest, amin_strided_batched_float_complex)
...
+    int            N            = std::get<0>(tup);
+    vector<double> alpha_beta   = std::get<1>(tup);
+    vector<int>    incx_incy    = std::get<2>(tup);
+    double         stride_scale = std::get<3>(tup);
+    int            batch_count  = std::get<4>(tup);
+    bool           fortran      = std::get<5>(tup);
+
+    Arguments arg;
+    arg.N      = N;
+    arg.alpha  = std::numeric_limits<double>::quiet_NaN();
```
TorreZuk, line 2433: Okay, if this matches blas.  I thought when alpha == 0 we fill with Nans to ensure the data isn't referenced.  Not clear the pattern you will follow but can wait and see.

TorreZuk (approved):
Will you be changing the default bench initializer?   This is more than just Nan work.

clients/gtest/blas1_gtest.cpp
daineAMD: You're right - I don't think this test adds much value, especially since we have the testing that you mentioned, and we don't have check_numerics to test here. I can get rid of these tests.
I also added the initialization work that Naveen added in rocBLAS out of convenience. I can move forward with those changes I think. We default to hpl initilization for rocBLAS bench, I'll change hipBLAS to match that in the next commit unless there's a reason to leave it with rand_int.

PR #451, opened 2022-01-26
Remove using namespace std

daineAMD:
- In `clients/gtest/\*\_gtest.cpp` replaced `using namespace std;` with `using std::vector;`
- In `clients/include/*.hpp`, removed `using namespace std;`
- Added `std::` to `vector` in three files (testing_trsm[_strided_batched].hpp)


TorreZuk (approved):
LGTM

PR #452, opened 2022-01-28
Updating init for L1 tests

daineAMD:


clients/include/testing_rot_ex.hpp
```diff
@@ -75,16 +75,15 @@ hipblasStatus_t testing_rot_ex_template(const Arguments& arg)
...
-
     // Random alpha (0 - 10)
     host_vector<int> alpha(1);
-    hipblas_init<int>(alpha, 1, 1, 1);
 
-    hipblas_init<Tcs>(hc, 1, 1, 1);
-    hipblas_init<Tcs>(hs, 1, 1, 1);
+    hipblas_init_vector(hx_host, arg, N, abs_incx, 0, 1, hipblas_client_alpha_sets_nan, true);
+    hipblas_init_vector(hy_host, arg, N, abs_incy, 0, 1, hipblas_client_alpha_sets_nan, false);
+    hipblas_init_vector(alpha, arg, 1, 1, 0, 1, hipblas_client_alpha_sets_nan, false);
```
TorreZuk, line 83: I think alpha would use never_sets_nan, it is only for testing alpha=0 like tests that would non even reference x & y

TorreZuk: Maybe they should all be never as you don't use alpha as rotation angle anymore?

clients/include/testing_rot_strided_batched.hpp
```diff
@@ -72,13 +72,15 @@ hipblasStatus_t testing_rot_strided_batched(const Arguments& arg)
...
 
     // Random alpha (0 - 10)
     host_vector<int> alpha(1);
-    hipblas_init<int>(alpha, 1, 1, 1);
+
+    hipblas_init_vector(
+        hx, arg, N, abs_incx, stride_x, batch_count, hipblas_client_alpha_sets_nan, true);
+    hipblas_init_vector(
+        hy, arg, N, abs_incy, stride_y, batch_count, hipblas_client_alpha_sets_nan, false);
+    hipblas_init_vector(alpha, arg, 1, 1, 0, 1, hipblas_client_alpha_sets_nan, false);
```
TorreZuk, line 83: same here, not sure if Nan is ever useful

TorreZuk (commented):
Basics look good, Naveen should review.  I think the best enum is not being used, normally only setting Nans for values that won't be used by kernel due to parameter values.  May be wrong if you are testing completely differently.

clients/include/testing_rot_ex.hpp
NaveenElumalaiAMD: Almost all Level 1 functions in rocBLAS had NaN initialization in the pre_checkin test and most of the functions neither have `alpha` nor `beta` parameters. Therefore, added  `hipblas_client_alpha_sets_nan` to initialize vectors with NaN's. I can also add another enum in `rocblas_check_nan_init` for Level 1 NaN initialization if this is misleading.

clients/include/testing_rotg.hpp
```diff
@@ -32,11 +32,10 @@ hipblasStatus_t testing_rotg(const Arguments& arg)
     host_vector<T> hs(1);
 
     // Initial data on CPU
-    srand(1);
-    hipblas_init<T>(ha, 1, 1, 1);
-    hipblas_init<T>(hb, 1, 1, 1);
-    hipblas_init<U>(hc, 1, 1, 1);
-    hipblas_init<T>(hs, 1, 1, 1);
+    hipblas_init_vector(ha, arg, 1, 1, 0, 1, hipblas_client_alpha_sets_nan, true);
```
NaveenElumalaiAMD, line 35: A very minor changez, could `arg` be renamed to `argus`? This would make it similar to all the other hipBLAS level 1 functions. Can be done in a separate PR.

NaveenElumalaiAMD (approved):
LGTM

clients/include/testing_rotg.hpp
daineAMD: It would be good to change this across the testing, there are `arg`s and `argus`s sprinkled throughout the files. I'll make a PR addressing this after these changes go in for L1, L2, and L3.

clients/include/testing_rot_ex.hpp
daineAMD: I see now that rocBLAS skips the NaN checks in this case. I think it's best to remove `hipblas_client_alpha_sets_nan` for the rot* cases for now, even though NaN tests haven't been added to the test suite yet.

PR #453, opened 2022-02-01
User specified initialization for L2

daineAMD:


TorreZuk (approved):
LGTM, should probably make a ticket for the init helpers to honor non absolute value increments, if not quick failure then we should both init and validate the negative incx,y etc.   Not sure what range are currently tested.  The same is true for rocblas.

PR #456, opened 2022-02-01
User specified initialization for L3

daineAMD:


TorreZuk:
This is strange as my PR-1068 on rocblas showed failures in trsm trsv gemm trmm all of a sudden on windows even though didn't expect they were affected.   Don't see how this PR affected the failures I see either

daineAMD:
Ah, that is odd. Must be related, these changes will probably continue to fail then.

TorreZuk:
Pulling the latest hip to see if I can reproduce locally.  Unfortunately the asum issue had been blocking other tests but this makes no sense to me now.   I wouldn't start changing tests to try and work around this, but if old initialization works that is also good to know.

daineAMD:
Passed on windows this time. Not sure if it's a fluke or if I happened to be working on the exact same thing in hipBLAS that happened to go wrong in rocBLAS.

TorreZuk (approved):
LGTM but would be good to know if batch 1000 trtri passes on another run

daineAMD:
All failures reappeared. Guessing the failures are intermittent and my changes aren't relevant.

TorreZuk:
> All failures reappeared. Guessing the failures are intermittent and my changes aren't relevant.

Force a re-run on windows with the same change.   Let's see if it passes, I am seeing the usual asum failure not occur on develop when I force a psdb test, but with no related changes it fails on PRs which do the psdb.   Could be something wrong with batches though so should keep digging a bit as if you repeat (with and without) and it fails and then passes again something is up.  Would also look at the artifacts it is using for the build, do they change?

PR #464, opened 2022-03-18
Update rocBLAS path windows

daineAMD:


cgmb (approved):

cgmb:
Looks like that matches the hipSOLVER PR: https://github.com/ROCmSoftwarePlatform/hipSOLVER/pull/87

TorreZuk (approved):
LGTM if it passes.   This and solver I seem to put it into staging/library but in the future might be better to be put into clients/staging, but maybe we'll drop and just rely on path instead.

TorreZuk:
Not so lucky. looks like ROCBLAS_PATH is too specific, now packages are installed generic to hipSDK so C:\hipSDK will be rocblas path and rocsolver path.  If CI isn't passing the arg to rmake.py its "default="C:/hipSDK/rocblas" needs to drop rocblas and the same for rocsolver.

cgmb:
> Not so lucky. looks like ROCBLAS_PATH is too specific, now packages are installed generic to hipSDK so C:\hipSDK will be rocblas path and rocsolver path. If CI isn't passing the arg to rmake.py its "default="C:/hipSDK/rocblas" needs to drop rocblas and the same for rocsolver.

I don't understand what you're referring to. However, I did notice that there's [a define in rocBLAS that doesn't appear to have been updated](https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/e4d904fa929c4f8c7e7c42eb4a5c246f45b652f6/library/src/tensile_host.cpp#L44).

daineAMD:
Windows CI built successfully, tests were aborted it looks like. I'll go ahead with the merge in a couple of minutes unless I hear otherwise.

TorreZuk:
Sounds good.  I think tensile init should have happened but it wasn't clear.  For runtime cloning into directory with rocblas.dll would like to see some gemms so re-ran the pipeline.

daineAMD:
Everything passed on the last run, so merging now.

PR #470, opened 2022-03-29
Updating documentation

daineAMD:
Updating based off [rocblas-internal#1153](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/1153).

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,18 +2,11 @@
 
 **hipBLAS** is a **BLAS** marshalling library, with multiple supported backends.  It sits between the application and a 'worker' BLAS library, marshalling inputs into the backend library and marshalling results back to the application.  **hipBLAS** exports an interface that does not require the client to change, regardless of the chosen backend.  Currently, **hipBLAS** supports **rocBLAS** and **cuBLAS** as backends.
 
-
-| Acronym  | Expansion                                                   |
-| :------- | :---------------------------------------------------------- |
-| **BLAS** | **B**asic **L**inear **A**lgebra **S**ubprograms            |
-| **ROCm** | **R**adeon **O**pen E**C**osyste**m**                       |
-| **HIP**  | **H**eterogeneous-Compute **I**nterface for **P**ortability |
-
 ## Documentation
 
 For a detailed description of the **hipBLAS** library, its implemented routines, the installation process and user guide, see the [**hipBLAS** Documentation](https://hipblas.readthedocs.io/en/latest/).
 
-hipBLAS requires either **rocBLAS** + **rocSOLVER** or **cuBLAS** APIs for BLAS implementation. For more information dependent **roc*** libraries see [rocBLAS documentation](https://rocblas.readthedocs.io/en/latest/), and [rocSolver documentation](https://rocsolver.readthedocs.io/en/latest/).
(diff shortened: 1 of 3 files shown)


TorreZuk (approved):
LGTM if links work.  Not sure why the other two acronyms are not kept, best practice is that you should define an acronym on first usage.   See https://github.com/ROCmSoftwarePlatform/rocBLAS-Examples

daineAMD:
The BLAS acronym was removed from rocBLAS in https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/970, not sure of the reason. I can put both HIP and BLAS back if we want, we should do the same in rocBLAS unless there's a reason not to.

README.md
```diff
@@ -2,18 +2,11 @@
...
-| **BLAS** | **B**asic **L**inear **A**lgebra **S**ubprograms            |
-| **ROCm** | **R**adeon **O**pen E**C**osyste**m**                       |
-| **HIP**  | **H**eterogeneous-Compute **I**nterface for **P**ortability |
-
 ## Documentation
 
 For a detailed description of the **hipBLAS** library, its implemented routines, the installation process and user guide, see the [**hipBLAS** Documentation](https://hipblas.readthedocs.io/en/latest/).
 
-hipBLAS requires either **rocBLAS** + **rocSOLVER** or **cuBLAS** APIs for BLAS implementation. For more information dependent **roc*** libraries see [rocBLAS documentation](https://rocblas.readthedocs.io/en/latest/), and [rocSolver documentation](https://rocsolver.readthedocs.io/en/latest/).
+hipBLAS requires either **rocBLAS** + **rocSOLVER** or **cuBLAS** APIs for BLAS implementation. For more information dependent **roc*** libraries see [rocBLAS and rocSOLVER documentation](https://docs.amd.com/category/libraries).
```
cgmb, line 9: I would wait to change these to docs.amd.com until the library documentation on the new site is a little more complete. https://docs.amd.com/bundle/RocBLAS_documentation/page/functions.html#rocblas-int

docs/source/install.rst
```diff
@@ -7,13 +7,13 @@ Prerequisites
...
 
-  * A ROCm enabled platform, more information `ROCm Documentation <https://rocm.github.io/>`_.
+  * A ROCm enabled platform, more information `ROCm Documentation <https://docs.amd.com/>`_.
   * A compatible version of rocBLAS
   * A compatible version of rocSOLVER for full functionality
 
 * If using the cuBLAS backend on a Nvidia machine:
 
-  * A HIP enabled platform, more information `HIP installation <https://rocmdocs.amd.com/en/latest/Installation_Guide/HIP-Installation.html#nvidia-platform/>`_.
+  * A HIP enabled platform, more information `HIP installation <https://docs.amd.com/bundle/HIP_Installation/page/HIP_Installation.html/>`_.
```
cgmb, line 16: You should drop the trailing slash. Our docs hosting website seems to be very forgiving of this sort of mistake, but it's kinda weird.

The id attribute of the "NVIDIA platform" heading on the new page is `d2e131`, so you should be able to link directly to it with https://docs.amd.com/bundle/HIP_Installation/page/HIP_Installation.html#d2e131.

README.md
daineAMD: Changed back for now.

PR #475, opened 2022-04-13
Cuda install flags

daineAMD:
- removed installation of cuda when passing in `-d` to install script. We aren't installing rocBLAS/rocSOLVER when passing in this flag either. It seems dangerous to do this as it will update cuda, even if it is already installed - users may not want this
- adds `--cudapath` to allow user to specify which cuda build they want to use (default to `/usr/local/cuda`. I've tested with a couple `/usr/local/cuda-VER` and it seems to work as expected)
- adds `--installcuda` to install cuda package via package manager, simply `apt install cuda` or similar
- adds `--installcudaversion` to be used with `--installcuda` to specify which version of cuda to install, for example: `./install.sh --cuda --installcuda --installcudaversion 11.4` will install cuda v11.4 using package manager

Tested flags on centos8 and ubuntu and they seem to work as expected. Working on testing sles now. Adding Cory as a reviewer as him and Torre always catch my cmake/bash errors :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,7 +73,7 @@ endif( )
 
 # Append our library helper cmake path and the cmake path for hip (for convenience)
 # Users may override HIP path by specifying their own in CMAKE_MODULE_PATH
-list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake  ${ROCM_PATH}/lib/cmake/hip /opt/rocm/lib/cmake/hip ${HIP_PATH}/cmake)
+list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake  ${ROCM_PATH}/lib/cmake/hip /opt/rocm/lib/cmake/hip ${HIP_PATH}/cmake ${CUDA_PATH})
 
 # NOTE:  workaround until hip cmake modules fixes symlink logic in their config files; remove when fixed
 list( APPEND CMAKE_PREFIX_PATH /opt/rocm /opt/rocm/llvm /opt/rocm/hip )
@@ -92,7 +92,8 @@ option( BUILD_SHARED_LIBS "Build hipBLAS as a shared library" ON )
 # Find CUDA if the user wants a CUDA version.
 option(USE_CUDA "Look for CUDA and use that as a backend if found" OFF)
 if (USE_CUDA)
-    find_package( CUDA REQUIRED )
+	set (ENV{CUDA_BIN_PATH} "${CUDA_PATH}" )
+	find_package( CUDA REQUIRED )
 endif()
 
 # Hip headers required of all clients; clients use hip to allocate device memory
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,7 +73,7 @@ endif( )
 
 # Append our library helper cmake path and the cmake path for hip (for convenience)
 # Users may override HIP path by specifying their own in CMAKE_MODULE_PATH
-list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake  ${ROCM_PATH}/lib/cmake/hip /opt/rocm/lib/cmake/hip ${HIP_PATH}/cmake ${CUDA_PATH})
+list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake  ${ROCM_PATH}/lib/cmake/hip /opt/rocm/lib/cmake/hip ${HIP_PATH}/cmake )
 
 # NOTE:  workaround until hip cmake modules fixes symlink logic in their config files; remove when fixed
 list( APPEND CMAKE_PREFIX_PATH /opt/rocm /opt/rocm/llvm /opt/rocm/hip )
@@ -92,7 +92,6 @@ option( BUILD_SHARED_LIBS "Build hipBLAS as a shared library" ON )
 # Find CUDA if the user wants a CUDA version.
 option(USE_CUDA "Look for CUDA and use that as a backend if found" OFF)
 if (USE_CUDA)
-	set (ENV{CUDA_BIN_PATH} "${CUDA_PATH}" )
 	find_package( CUDA REQUIRED )
 endif()
 
--- a/install.sh
+++ b/install.sh
@@ -455,7 +455,7 @@ declare -a cmake_client_options
 # check if we have a modern version of getopt that can handle whitespace and long parameters
 getopt -T
 if [[ $? -eq 4 ]]; then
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -73,7 +73,7 @@ endif( )
 
 # Append our library helper cmake path and the cmake path for hip (for convenience)
 # Users may override HIP path by specifying their own in CMAKE_MODULE_PATH
-list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake  ${ROCM_PATH}/lib/cmake/hip /opt/rocm/lib/cmake/hip ${HIP_PATH}/cmake)
+list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake  ${ROCM_PATH}/lib/cmake/hip /opt/rocm/lib/cmake/hip ${HIP_PATH}/cmake ${CUDA_PATH})
```
cgmb, line 76: Is there a particular error condition that adding `CUDA_PATH` to the `CMAKE_MODULE_PATH` is intended to resolve? I wouldn't expect this to be necessary.

CMakeLists.txt
```diff
@@ -92,7 +92,8 @@ option( BUILD_SHARED_LIBS "Build hipBLAS as a shared library" ON )
 # Find CUDA if the user wants a CUDA version.
 option(USE_CUDA "Look for CUDA and use that as a backend if found" OFF)
 if (USE_CUDA)
-    find_package( CUDA REQUIRED )
+	set (ENV{CUDA_BIN_PATH} "${CUDA_PATH}" )
+	find_package( CUDA REQUIRED )
```
cgmb, line 96: I would suggest leaving this bit unchanged and setting `CUDA_BIN_PATH` via `install.sh`. That way, experienced users of CMake who are familiar with the standard ways of customizing the search paths for CUDA will be able to build using CMake directly using options they're already familiar with.

You actually do already set `CUDA_BIN_PATH` in the install script, but I think you may have forgotten to export it.

install.sh
```diff
@@ -414,6 +511,16 @@ while true; do
     --cuda|--use-cuda)
         build_cuda=true
         shift ;;
+    --cudapath)
+	cuda_path=${2}
+	CUDA_BIN_PATH=${cuda_path}
```
cgmb, line 516: Not all bash variables are inherited by the child processes of the script. Only exported variables are visible to child processes. This defines `CUDA_BIN_PATH` for `install.sh`, but it won't be visible in your `cmake` calls unless you add `export CUDA_BIN_PATH`. Since this variable is not exported and is not referenced anywhere else in the script, it has no effect.

Alternatively, you could move `CUDA_BIN_PATH=${cuda_path}` to be a variable specified just for the CMake invokation, like `CXX=${compiler}` is.

install.sh
```diff
@@ -616,13 +727,13 @@ pushd .
...
     CXX=${compiler} ${cmake_executable} ${cmake_common_options[@]} ${cmake_client_options[@]} -DCPACK_SET_DESTDIR=OFF -DCMAKE_INSTALL_PREFIX="${rocm_path}" \
-    -DCMAKE_PREFIX_PATH="${rocm_path};${rocm_path}/hcc;${rocm_path}/hip;$(pwd)/../deps/deps-install;/usr/local/cuda;${cmake_prefix_path}" \
+    -DCMAKE_PREFIX_PATH="${rocm_path};${rocm_path}/hcc;${rocm_path}/hip;$(pwd)/../deps/deps-install;${cuda_path};${cmake_prefix_path}" \
     -DCMAKE_SHARED_LINKER_FLAGS="${rocm_rpath}" \
     -DCMAKE_EXE_LINKER_FLAGS=" -Wl,--enable-new-dtags -Wl,--rpath,${rocm_path}/lib:${rocm_path}/lib64" \
     -DROCM_DISABLE_LDCONFIG=ON \
     -DROCM_PATH="${rocm_path}" ../..
   else
-    CXX=${compiler} ${cmake_executable} ${cmake_common_options[@]} ${cmake_client_options[@]} -DCPACK_SET_DESTDIR=OFF -DCMAKE_PREFIX_PATH="$(pwd)/../deps/deps-install;${cmake_prefix_path}" -DROCM_PATH=${rocm_path} ../..
+    CXX=${compiler} ${cmake_executable} ${cmake_common_options[@]} ${cmake_client_options[@]} -DCPACK_SET_DESTDIR=OFF -DCMAKE_PREFIX_PATH="$(pwd)/../deps/deps-install;${cmake_prefix_path}" -DROCM_PATH=${rocm_path} -DCUDA_PATH=${cuda_path} ../..
```
cgmb, line 736: If fixing the export of `CUDA_BIN_PATH` allows you to drop the changes to CMakeLists.txt, you'd be able to drop the `-DCUDA_PATH` from this command as well.

install.sh
```diff
@@ -285,6 +323,51 @@ install_packages( )
...
+
+  local cuda_dependencies=("cuda")
+  case "${ID}" in
+	  ubuntu)
+		  elevate_if_not_root apt update
+		  if [[ "${cuda_version_install}" == "default" ]]; then
+			  install_apt_packages "${cuda_dependencies[@]}"
+	          else
+		 	  install_apt_packages_version "${cuda_dependencies[@]}" "${cuda_version_install}"
+                  fi
```
cgmb, line 341: There's some mangled formatting here.

install.sh
```diff
@@ -285,6 +323,51 @@ install_packages( )
...
+		  else
+		  	  install_dnf_packages_version "${cuda_dependencies[@]}" "${cuda_version_install}"
+		  fi
+		  ;;
+	  sles|opensuse-leap)
+		  if [[ "${cuda_version_install}" == "default" ]]; then
+			  install_zypper_packages "${cuda_dependencies[@]}"
+		  else
+		  	install_zypper_packages_version "${cuda_dependencies[@]}" "${cuda_version_install}"
+	          fi
```
cgmb, line 362: And here.

install.sh
```diff
@@ -35,7 +37,11 @@ cat <<EOF
     --codecoverage                Build with code coverage profiling enabled, excluding release mode.
 
     -d, --dependencies            Build and install external dependencies. Dependecies are to be installed in /usr/local.
-                                  This should be done only once (this does not install rocBLAS or rocSolver).
+                                  This should be done only once (this does not install rocBLAS, rocSolver, or cuda).
+
+    --installcuda                 Install cuda pacakge
```
cgmb, line 42: `s/pacakge/package.`

install.sh
```diff
@@ -297,6 +380,17 @@ install_zypper_packages( )
...
 
+install_zypper_packages_version( )
+{
+  package_dependencies=("$1")
+  package_versions=("$2")
+  for index in ${package_dependencies[*]}; do
+    printf "\033[32mInstalling \033[33m${package_dependencies[$index]} version ${package_versions[$index]} from distro package manage
+r \033[0m\n"
+    elevate_if_not_root zypper -n --no-gpg-checks install ${package_dependencies[$index]}-${package_versions[$index]}
+  done
```
cgmb, line 391: Is there a reason we do these one at a time? I know that's how the existing installs are done, but it's way faster to install everything with one call to `zypper/apt/yum/dnf`.

cgmb (commented):
This looks pretty handy! Though, I think you could probably get it working without making any changes to the existing CMake code. If it worked without changes, that would be preferable.

install.sh
daineAMD: Not that I know of, I created LWPMLSE-243 to keep PRs focused.

cgmb:
Assuming you've tested these options and they work as expected, LGTM.

cgmb (approved):

PR #478, opened 2022-04-27
Installing dependencies in one call

daineAMD:
Same changes as in https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/1204

- Didn't change the `install_pkgmanager_packages_version()` method I introduced previously as it's only used to install `cuda` for now, can update in the future if needed.
- Added gcc-fortran dependency to sles as it's needed
- Ran into the same devtoolset-7-gcc-gfortran / gcc-gfortran dependency issue on centos7 as mentioned in the rocBLAS PR, can discuss in a meeting

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -283,7 +283,7 @@ install_packages( )
     library_dependencies_centos+=( "devtoolset-7-gcc-gfortran" )
     library_dependencies_centos8+=( "gcc-gfortran" )
     library_dependencies_fedora+=( "gcc-gfortran" )
-    library_dependencies_sles+=( "pkg-config" "dpkg" )
+    library_dependencies_sles+=( "gcc-fortran pkg-config" "dpkg" )
   fi
 
   case "${ID}" in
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -259,27 +271,25 @@ install_packages( )
...
 
-  local client_dependencies_ubuntu=( "gfortran" )
-  local client_dependencies_centos=( "devtoolset-7-gcc-gfortran" )
-  local client_dependencies_centos8=( "gcc-gfortran" )
-  local client_dependencies_fedora=( "gcc-gfortran" )
-  local client_dependencies_sles=( "pkg-config" "dpkg" )
+  if [[ "${build_clients}" == true ]]; then
+    library_dependencies_ubuntu+=( "gfortran" )
+    library_dependencies_centos+=( "devtoolset-7-gcc-gfortran" )
+    library_dependencies_centos8+=( "gcc-gfortran" )
```
TorreZuk, line 284: So rhel and centos are using identical deps any reason you didn't want to call these centos_rhel as before?

TorreZuk (approved):

install.sh
TorreZuk: Well they may diverge with stream if not 8.6 so doesn't hurt to be clearer it is used for both.

PR #480, opened 2022-04-28
Reducing test time

daineAMD:
Removing some tests from hipBLAS, these changes make it take about half the time on my vega20 compared to previously. Windows tests have been taking a long time, this is some low hanging fruit to try to reduce that.

This is mostly reducing batched tests which took longer than expected. We don't need to do exhaustive testing here because the backends themselves should ensure their own correctness.

See LWPMLSE-232 for more info on test timing.

### DIFF
--- a/clients/gtest/gbmv_batched_gtest.cpp
+++ b/clients/gtest/gbmv_batched_gtest.cpp
@@ -41,28 +41,20 @@ Representative sampling is sufficient, endless brute-force sampling is not neces
 // add/delete as a group
 const vector<vector<int>> matrix_size_range = {
     {-1, -1, -1, -1, -1},
-    //        {10, 10, 2},
-    //        {600,500, 500},
     {1000, 1000, 150, 84, 1000},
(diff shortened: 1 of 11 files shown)


clients/gtest/gbmv_batched_gtest.cpp
```diff
@@ -72,11 +64,8 @@ const vector<char> transA_range = {
     // 'C',
 };
 
-// number of gemms in batched gemm
-const vector<int> batch_count_range = {
```
TorreZuk, line 76: You could remove size 1 and change 5 to batch 2 if you really just want to ensure the looping over batches.  This is what I set for some bad_arg batch sizes in rocblas, just ensure a hard coded 1 doesn't get used somewhere.

TorreZuk (approved):

daineAMD:
Windows tests took 30 mins as opposed to develop's 2h30m+ last run, so hopefully that reduction is consistent. CI failures due to downtime, saw passes otherwise.

PR #482, opened 2022-05-03
Adding surface plot to performance scripts

daineAMD:
- Adding surface_plot [as done in rocBLAS](https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/841).
- Made some other small changes to make python files more generic as they can be shared between rocBLAS and hipBLAS in the future, or other libraries if applicable. I'm not sure on the future of these in regards to performance measurement vs. the new performance tracking tool in the works, so didn't want to spend too much time on this right now in case we don't need it in the future


NaveenElumalaiAMD (approved):
Checked out the changes, the new surface plot feature and the normal plot works well!

PR #483, opened 2022-05-09
Adding a script to plot multiple performance runs together

daineAMD:
Adding a script which can be used with performance data from performancereport.py to plot the performance of multiple runs of different functions together. See LWPMLSE-253 for some example plots created with this script.

Example usage:

`./combine_plots.py -d output1/run00 -d output2/run00`
where output1 and output2 are directories containing the output from a performancereport.py run. Or the directories could be manually created with just the input/output from hipblas-bench command line runs.


TorreZuk (approved):
LGTM

PR #485, opened 2022-05-21
for RHEL8 use MAJORVERSION in place of VERSION_ID

amcamd:
- resolves [SWDEV-336664](https://ontrack-internal.amd.com/browse/SWDEV-336664)
- the bash shell compare needs an integer, not a floating point number
- use the major version which is an integer in place of the version_id


amcamd:
Close because #486 fix is better

PR #505, opened 2022-07-07
Remove all references to hcc

cgmb:
- Match HIP platform naming convention:
  hcc  -> amd
  nvcc -> nvidia
- Remove old references to hcc and replace with hipcc as needed.


daineAMD (approved):
Cool, thanks for doing this! Not sure if anyone would have a preference of `amd_detail`/`nvidia_detail` over something like `rocm_detail`/`cuda_detail`.

PR #506, opened 2022-07-07
Adding hipblasXgels

daineAMD:
Adding hipblasXgels for [SWDEV-344553](https://ontrack-internal.amd.com/browse/SWDEV-344553).

~~Still need to add gflop calculations.~~ Also I'm not familiar with gels so extra review by someone from rocSOLVER in testing_gels.hpp would be appreciated.

Will add batched and stridedbatched variants in another PR, trying to keep them easier to review.

### DIFF
--- a/clients/gtest/gels_gtest.cpp
+++ b/clients/gtest/gels_gtest.cpp
@@ -90,7 +90,7 @@ TEST_P(gels_gtest, gels_gtest_float)
     if(status != HIPBLAS_STATUS_SUCCESS)
     {
         if(arg.M < 0 || arg.N < 0 || arg.K < 0 || arg.lda < arg.M || arg.ldb < arg.M
-           || arg.ldb < arg.N || arg.transA_option == 'C')
+           || arg.ldb < arg.N)
         {
             EXPECT_EQ(HIPBLAS_STATUS_INVALID_VALUE, status);
         }
@@ -113,7 +113,7 @@ TEST_P(gels_gtest, gels_gtest_double)
     if(status != HIPBLAS_STATUS_SUCCESS)
     {
         if(arg.M < 0 || arg.N < 0 || arg.K < 0 || arg.lda < arg.M || arg.ldb < arg.M
-           || arg.ldb < arg.N || arg.transA_option == 'C')
+           || arg.ldb < arg.N)
         {
             EXPECT_EQ(HIPBLAS_STATUS_INVALID_VALUE, status);
         }
@@ -136,7 +136,7 @@ TEST_P(gels_gtest, gels_gtest_float_complex)
     if(status != HIPBLAS_STATUS_SUCCESS)
     {
         if(arg.M < 0 || arg.N < 0 || arg.K < 0 || arg.lda < arg.M || arg.ldb < arg.M
-           || arg.ldb < arg.N || arg.transA_option == 'T')
+           || arg.ldb < arg.N)
         {
             EXPECT_EQ(HIPBLAS_STATUS_INVALID_VALUE, status);
         }
@@ -159,7 +159,7 @@ TEST_P(gels_gtest, gels_gtest_double_complex)
     if(status != HIPBLAS_STATUS_SUCCESS)
     {
         if(arg.M < 0 || arg.N < 0 || arg.K < 0 || arg.lda < arg.M || arg.ldb < arg.M
-           || arg.ldb < arg.N || arg.transA_option == 'T')
+           || arg.ldb < arg.N)
         {
             EXPECT_EQ(HIPBLAS_STATUS_INVALID_VALUE, status);
         }
--- a/clients/gtest/gels_gtest.cpp
+++ b/clients/gtest/gels_gtest.cpp
@@ -76,7 +76,7 @@ protected:
     virtual void TearDown() {}
 };
 
-// #ifndef __HIP_PLATFORM_NVCC__
+#ifndef __HIP_PLATFORM_NVCC__
 
 TEST_P(gels_gtest, gels_gtest_float)
 {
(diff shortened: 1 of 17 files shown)


amcamd (approved):
LGTM

library/src/hcc_detail/hipblas.cpp
```diff
@@ -16627,6 +16628,122 @@ catch(...)
...
+hipblasStatus_t hipblasSgels(hipblasHandle_t    handle,
+                             hipblasOperation_t trans,
+                             const int          m,
+                             const int          n,
+                             const int          nrhs,
+                             float*             A,
+                             const int          lda,
+                             float*             B,
+                             const int          ldb,
+                             int*               info)
```
tfalders, line 16640: I'd suggest making the method signature closer to what cuBLAS has for gelsBatched, which in this case is an additional info pointer on the host. It would be nice if you could set the value of this pointer similar to how it's done by hipBLAS's getrs and geqrf methods.

You might also want to consider only adding the batched method, like how we did for getri. Up to you.

library/src/hipblas_module.f90
```diff
@@ -13236,4 +13236,85 @@ function hipblasZgeqrfStridedBatched(handle, m, n, A, lda, stride_A,&
         end function hipblasZgeqrfStridedBatched
     end interface
 
+    ! gels
+    interface
+        function hipblasSgels(handle, m, n, nrhs, trans, A, lda, B, ldb, info) &
+                result(c_int) &
+                bind(c, name = 'hipblasSgels')
```
tfalders, line 13243: @cgmb brought to my attention that this is the wrong way to declare a Fortran result value. It goes beyond the scope of this PR, but you might want to consider making the same change as we made in hipSOLVER: https://github.com/ROCmSoftwarePlatform/hipSOLVER/pull/105.

clients/gtest/gels_gtest.cpp
```diff
@@ -0,0 +1,184 @@
...
+typedef std::tuple<vector<int>, char, bool> gels_tuple;
+
+// {m, n, nrhs, lda, ldb}
+const vector<vector<int>> matrix_size_range
+    = {{-1, -1, -1, 1, 1}, {10, 10, 10, 10, 10}, {10, 10, 10, 20, 100}, {600, 500, 400, 600, 600}};
+
+const vector<char> trans_range = {
+    'N',
+    'T',
+};
```
tfalders, line 45: Note that cuBLAS's gels only supports non-transposed and overdetermined cases, so it might be worth removing the transpose test.

tfalders (commented):
Looks pretty good! I do have a few concerns, though. See comments.

library/src/hipblas_module.f90
TorreZuk: Yes not in this PR but we should make these fixes, there was an email chain where it was suggested we were going to pass this off to hipFortran but we should just go ahead and fix if you fixed solver.  Need to fix rocblas as well, I can do rocblas Daine if you will fix hipblas.

library/include/hipblas.h
```diff
@@ -16634,6 +16634,113 @@ HIPBLAS_EXPORT hipblasStatus_t hipblasZgetriBatched(hipblasHandle_t
...
+    n           int. n >= 0.\n
+                The number of columns of matrix A.
+    @param[in]
+    nrhs        int. nrhs >= 0.\n
+                The number of columns of matrices B and X;
+                i.e., the columns on the right hand side.
+    @param[inout]
+    A           pointer to type. Array on the GPU of dimension lda*n.\n
+                On entry, the matrix A.
+                On exit, the QR (or LQ) factorization of A as returned by \ref rocsolver_sgeqrf "GEQRF" (or \ref rocsolver_sgelqf "GELQF").
```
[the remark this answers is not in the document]
cgmb: We _might_ be able to do it if we manually author the [Intersphinx](https://docs.readthedocs.io/en/stable/guides/intersphinx.html) references, but it certainly won't work automatically.

library/src/hipblas_module.f90
TorreZuk: Yes would't change just one function, we can look if we ever made a ticket, if not one lwp ticket for both hipblas and rocblas can suffice I would think.

library/include/hipblas.h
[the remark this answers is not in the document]
daineAMD: Removed the reference and just left the name of the function in place.

clients/gtest/gels_gtest.cpp
daineAMD: Left for this one, I'll keep this in mind when adding the batched version, though.

library/src/hcc_detail/hipblas.cpp
tfalders: My understanding is that the two info pointers communicate different information. The first info is on the host and returns a value <= 0 depending on the presence of invalid arguments. The second info is on the device and returns a value >= 0 depending on singularities in the problem.

daineAMD: Ah, my fault, I see the difference now, yes I'll update this and let you know.

daineAMD: Added, haven't tested myself so will wait for CI. Let me know if the arg checking in hcc_detail/hipblas.cpp looks correct.

library/include/hipblas.h
```diff
@@ -16693,7 +16693,11 @@ HIPBLAS_EXPORT hipblasStatus_t hipblasZgetriBatched(hipblasHandle_t
     ldb         int. ldb >= max(m,n).\n
                 Specifies the leading dimension of matrix B.
     @param[out]
-    info        pointer to int on the GPU.\n
+    info      pointer to a int on the host.\n
+              If info = 0, successful exit.
+              If info = j < 0, the j-th argument is invalid.
```
tfalders, line 16698: Very minor, but it looks like the parameter description is misaligned with other parameters.

tfalders (approved):
LGTM! Just one minor comment. Thanks for taking care of this!

PR #507, opened 2022-07-12
update fortran for non gfortran

TorreZuk:
* Correction interface FORTRAN rocBLAS pour le compilateur Cray

### DIFF
--- a/clients/include/hipblas_fortran.f90
+++ b/clients/include/hipblas_fortran.f90
@@ -25,46 +25,49 @@ module hipblas_interface
     use iso_c_binding
     use hipblas
 
-    contains
+contains
 
     !--------!
     !  Aux   !
     !--------!
     function hipblasSetVectorFortran(n, elemSize, x, incx, y, incy) &
-            result(res) &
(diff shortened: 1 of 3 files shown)


TorreZuk:
Hi @daineAMD please review, maybe should add the git code owners file and settings so we don't have to add reviewers by default?

TorreZuk:
> LGTM after CI picks up the latest fix. Will create codeowners PR in an hour or two.

One func got missed but hopefully builds now. Is the fortran API tested like rocblas does?  _F in name shows fortran run?

daineAMD:
> > LGTM after CI picks up the latest fix. Will create codeowners PR in an hour or two.
> 
> One func got missed but hopefully builds now. Is the fortran API tested like rocblas does? _F in name shows fortran run?

Unfortunately no we don't really have any useful information in the naming scheme of tests in hipBLAS currently. You can browse around clients/gtest/*_gtest.cpp to see how the tests are created.

TorreZuk:
@daineAMD CI related failures only I could see so merged so you can rebase your PR as you had some overlap.

PR #508, opened 2022-07-12
Adding gelsBatched and gelsStridedBatched

daineAMD:


TorreZuk (approved):
LGTM if all tests pass, but will leave the solver reviewer to verify the usage.

So if most follow this test pattern then fortran API is tested for all other parameter permutations?  Seems like overkill but as testing is quick not a problem for now.

daineAMD:
> LGTM if all tests pass, but will leave the solver reviewer to verify the usage.
> 
> So if most follow this test pattern then fortran API is tested for all other parameter permutations? Seems like overkill but as testing is quick not a problem for now.

Yep, that's right. Test infrastructure could definitely be overhauled at some point. The suite is fairly small though so don't think it's too urgent as you say.

clients/common/hipblas_template_specialization.cpp
```diff
@@ -20427,4 +20581,158 @@ hipblasStatus_t hipblasGels<hipblasDoubleComplex, true>(hipblasHandle_t       ha
...
+                                                               const int                   m,
+                                                               const int                   n,
+                                                               const int                   nrhs,
+                                                               hipblasDoubleComplex* const A[],
+                                                               const int                   lda,
+                                                               hipblasDoubleComplex* const B[],
+                                                               const int                   ldb,
+                                                               int*                        info,
+                                                               int*      deviceInfo,
+                                                               const int batchCount)
```
tfalders, line 20651: Clang-format doesn't seem to have aligned these properly.

clients/common/hipblas_template_specialization.cpp
```diff
@@ -20427,4 +20581,158 @@ hipblasStatus_t hipblasGels<hipblasDoubleComplex, true>(hipblasHandle_t       ha
...
+                                                                      const int             nrhs,
+                                                                      hipblasDoubleComplex* A,
+                                                                      const int             lda,
+                                                                      const hipblasStride   strideA,
+                                                                      hipblasDoubleComplex* B,
+                                                                      const int             ldb,
+                                                                      const hipblasStride   strideB,
+                                                                      int*                  info,
+                                                                      int*      deviceInfo,
+                                                                      const int batchCount)
```
tfalders, line 20732: Same here

clients/include/testing_gels_batched.hpp
```diff
@@ -0,0 +1,195 @@
...
+
+        hipblas_error
+            = norm_check_general<T>('F', std::max(M, N), nrhs, ldb, hB, hB_res, batchCount);
+
+        for(int b = 0; b < batchCount; b++)
+        {
+            if(info_res[b] != 0)
+                hipblas_error += 1.0;
+            if(info[b] != info_input)
+                hipblas_error += 1.0;
```
tfalders, line 148: I missed this on the first PR, but I think it makes more sense to compare `info[b]` with `info_res[b]`. If the matrix is singular, `info[b]` and `info_res[b]` will both be non-zero while `info_input` should be zero.

clients/include/testing_gels_strided_batched.hpp
```diff
@@ -0,0 +1,214 @@
...
+        hipblas_error = norm_check_general<T>(
+            'F', std::max(M, N), nrhs, ldb, strideB, hB.data(), hB_res.data(), batchCount);
+
+        for(int b = 0; b < batchCount; b++)
+        {
+            if(info_res[b] != 0)
+                hipblas_error += 1.0;
+
+            if(info[b] != info_input)
+                hipblas_error += 1.0;
```
tfalders, line 165: Same thing here

library/src/hcc_detail/hipblas.cpp
```diff
@@ -16832,6 +16841,480 @@ catch(...)
...
+    else if(A == NULL)
+        *info = -5;
+    else if(lda < m)
+        *info = -6;
+    else if(B == NULL)
+        *info = -7;
+    else if(ldb < m || ldb < n)
+        *info = -8;
+    else if(info == NULL)
+        *info = -9;
```
tfalders, line 16876: I suppose these checks on `info` can be removed, since they should be unreachable (and would segfault).

tfalders (commented):
Looks good! Just a few minor comments.

library/src/hcc_detail/hipblas.cpp
daineAMD: lol yeah, a pretty bad two lines of code just looking at it now.

clients/include/testing_gels_strided_batched.hpp
daineAMD: Ah you're right, fixed.

tfalders (approved):

PR #514, opened 2022-08-29
Updating conf.py copyright date

daineAMD:
As per Colin's comment in Teams

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -56,7 +56,7 @@ master_doc = 'index'
 
 # General information about the project.
 project = u'hipBLAS'
-copyright = u'2016-2021, Advanced Micro Devices'
+copyright = u'2016-2022, Advanced Micro Devices'
 author = u'Advanced Micro Devices'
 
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -56,7 +56,7 @@
 
 # General information about the project.
 project = u'hipBLAS'
-copyright = u'2016-2021, Advanced Micro Devices'
+copyright = u'2016-2022, Advanced Micro Devices'
 author = u'Advanced Micro Devices'
 
 # The version info for the project you're documenting, acts as replacement for
```
amcamd, line 62: does the version and release information below also need to be updated. it is currently at 0.8 I think it should be hipBLAS 0.52.0 for ROCm 5.3.0

daineAMD: Looks like they're just comments so shouldn't matter. I don't really have any context for this file besides it being related to docs/sphinx, though, so I'm not sure where the versions are handled, @rkamd might know.

amcamd (approved):
Colin confirmed the version can remain commented out

PR #517, opened 2022-09-08
Change dot_ex tests to use near_check for fp16

daineAMD:
Some dot_ex tests failed while testing out gfx11. It appears that it's just a conversion from fp16->fp32->fp16->fp32 which is causing the issue. Only seen when the result is 0, we get a cpu result of 0 and a gpu result of (very small number). I added some example output at SWDEV-333517.

This change dot_ex tests to use near_check rather than unit_check for half precision. Using a tolerance of epsilon * N.

### DIFF
--- a/clients/include/testing_dot_batched_ex.hpp
+++ b/clients/include/testing_dot_batched_ex.hpp
@@ -105,7 +105,7 @@ hipblasStatus_t testing_dot_batched_ex_template(const Arguments& argus)
     double gpu_time_used, hipblas_error_host, hipblas_error_device;
 
     // Initial Data on CPU
-    hipblas_init(hy, true, true);
+    hipblas_init(hy, true, false);
     hipblas_init_alternating_sign(hx);
     CHECK_HIP_ERROR(dx.transfer_from(hx));
     CHECK_HIP_ERROR(dy.transfer_from(hy));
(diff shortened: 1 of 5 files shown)


clients/include/testing_dot_batched_ex.hpp
```diff
@@ -105,7 +105,7 @@ hipblasStatus_t testing_dot_batched_ex_template(const Arguments& argus)
     double gpu_time_used, hipblas_error_host, hipblas_error_device;
 
     // Initial Data on CPU
-    hipblas_init(hy, true, true);
+    hipblas_init(hy, true, false);
```
TorreZuk, line 108: If you aren't alternating signs won't we have potential large N unit check failures?

clients/include/testing_dot_strided_batched_ex.hpp
```diff
@@ -180,8 +180,31 @@ hipblasStatus_t testing_dot_strided_batched_ex_template(const Arguments& argus)
 
         if(argus.unit_check)
         {
-            unit_check_general<Tr>(1, batch_count, 1, h_cpu_result, h_hipblas_result_host);
-            unit_check_general<Tr>(1, batch_count, 1, h_cpu_result, h_hipblas_result_device);
+            if(std::is_same<Tr, hipblasHalf>{})
+            {
+                double tol = pow(2, -14) * N;
```
TorreZuk, line 185: Not sure but a helper function may be better, which can have comment on usage validity.   Could be put in near.xxx

TorreZuk (approved):
Suggestions, are these new failures?   I would double check the random input values as the odds of the dot result being zero should be very low for half and other.

clients/include/testing_dot_batched_ex.hpp
daineAMD: hx has alternating input below this.

daineAMD:
The inputs are in the range int([1,3]) for y and int([-3, 3]) for x for half precision, so odds of a 0 result aren't super low. The failures are quite intermittent, only really pop in the strided_batched tests as there are more opportunities for 0 results.

TorreZuk:
> The inputs are in the range int([1,3]) for y and int([-3, 3]) for x for half precision, so odds of a 0 result aren't super low. The failures are quite intermittent, only really pop in the strided_batched tests as there are more opportunities for 0 results.

Okay yes so an integer random walk around zero, thought we used 10 as the range I'll have to re-read it.  And HPL is only benchmark initialization, anyway I approved.  I'll maybe plot this random walk to know better is going on for our random sequence.  Maybe should move to real [-1,1] initialization for testing too.

PR #518, opened 2022-09-09
Openblas reference code for windows

daineAMD:
Should be similar changes to in https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/913

Adding support for openblas on windows. Doesn't necessarily have to go in before feature-complete, just made it easier to test out gfx11 on windows.


clients/CMakeLists.txt
```diff
@@ -69,14 +69,50 @@ if( BUILD_CLIENTS_TESTS OR BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_SAMPLES )
...
+      set( BLAS_LIBRARY "blas" )
+    endif()
+  else() # WIN32
+    set( BLAS_INCLUDE_DIR ${OPENBLAS_DIR}/include CACHE PATH "OpenBLAS library include path" )
+    find_library( BLAS_LIBRARY libopenblas
+                  PATHS ${OPENBLAS_DIR}/lib
+                  REQUIRED
+                  NO_DEFAULT_PATH
+                )
+  endif()
```
TorreZuk, line 94: You should look at the current rocblas this is old code, it has a fallback to use OpenBLAS from vcpkg.  Then rdeps.py can install your dependencies on windows for hipblas as well....

rdeps.xml
```diff
@@ -0,0 +1,18 @@
...
+      <pkg name="google test">gtest:x64-windows</pkg>
+      <pkg name="OpenBLAS">openblas:x64-windows</pkg>
+    </vcpkg>
+  </os>
+  <os names="all">
+    <pip>
+      <pkg name="PyYAML">pyyaml</pkg>
+      <pkg name="six">six</pkg>
+      <pkg name="wheel">wheel</pkg>
+      <pkg name="virtualenv">virtualenv</pkg>
```
TorreZuk, line 15: Do you need wheel and virtualenv?

daineAMD: No, I should be able to get rid of them. Do you happen to know if `six` is used for something other than a virtualenv dependency in rocBLAS?

TorreZuk (approved):
If you can get all to pass I approve. Your error looks like it might be covered in my PR, if this is valid: hipblas_parse_data.cpp:(.text+0x3c7): undefined reference to `std::experimental::filesystem::v1::path::_M_split_cmpts()'
See PR #https://github.com/ROCmSoftwarePlatform/hipBLAS/pull/523

rdeps.xml
TorreZuk: If you grep for six and nothing shows I would try to remove it.

PR #523, opened 2022-09-16
Lwpmlse 298 bench yaml

TorreZuk:
* uses c++17 to standardize some code with rocblas
* adds yaml processing support for hipblas-bench --yaml <file>
* adds bfloat16 helpers into class, only used by clients (host side only code) which adds support to output alpha/beta values
* adds yaml processing of atomics mode
* fixes some warnings with rocm5.4 compiler


clients/common/hipblas_gentest.py
```diff
@@ -0,0 +1,574 @@
...
+        test[key] = int(result)
+
+
+def setdefaults(test):
+    """Set default values for parameters"""
+    # Do not put constant defaults here -- use hipblas_common.yaml for that.
+    # These are only for dynamic defaults
+    # TODO: This should be ideally moved to YAML file, with eval'd expressions.
+
+    if test['function'] in ('asum_strided_batched', 'nrm2_strided_batched',
```
daineAMD, line 229: It's a little strange that we're using underscores in function names when the functions are actually CamelCase. This is prevalent in other places in the hipBLAS clients, though, so maybe a future task to change them all.

daineAMD (approved):

clients/common/hipblas_gentest.py
TorreZuk: It is example code it won't be true unless we add test or bench yaml with to those names I to think.
We might consider no smart defaults at all.  This just adds the system, actual use will be another PR.  I thought maybe hipblas_smoke.yaml might be first use case?

TorreZuk:
All CI related failures so merged

PR #525, opened 2022-09-19
remove HIP_PATH in lieu of HIP_DIR

daineAMD:
I noticed that rocBLAS no longer uses HIP_PATH, only HIP_DIR. The [Windows wiki](https://confluence.amd.com/display/MLSE/MLSE+library+windows+support) mentions both HIP_PATH and HIP_DIR, but it seems that only HIP_DIR is needed in rocBLAS, and in hipBLAS after this PR.

Also removes some code that gave errors with windows slashes in the environment variable.


TorreZuk (approved):
Fine to match rocblas.  Might want to simplify further once we get the sdk ready but for now providing the path seems okay.

PR #528, opened 2022-09-21
fix earlier bad merge

TorreZuk:
* order needs to match definition


daineAMD (approved):
This just reverts the changes made in this file, correct?

TorreZuk:
> This just reverts the changes made in this file, correct?

No revert, just fixes the ordering which got messed.  No tests need this but yaml driven bench and test will require it

PR #530, opened 2022-09-28
Lwpmlse 329 hipblas test yaml

TorreZuk:
* support yaml driven testing alongside original testing framework, using snake case tokens for now
* adds --yaml hipblas_smoke.yaml
* fix general testing support found using new smoke test set
* unifies style of test framework with model and parameter to stack local var pattern
* lwpmlse-329


clients/include/hipblas_test.hpp
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    virtual ~hipblas_test_invalid() = default;
+};
+
+/* ============================================================================================ */
+/*! \brief  Normalized test name to conform to Google Tests */
+// The template parameter is only used to generate multiple instantiations with distinct static local variables
+template <typename>
+class RocBLAS_TestName
```
daineAMD, line 97: A couple instances of rocblas remaining instead of hipblas.

daineAMD (commented):
About halfway done reviewing, looks good so far - will finish review tomorrow.

clients/include/hipblas_test.hpp
TorreZuk: Thanks that may be dead code, will fix.  Sorry for the PR size it just kept growing 🤣

clients/include/testing_proto.hpp
```diff
@@ -0,0 +1,623 @@
...
+// Copyright (C) 2022 Advanced Micro Devices, Inc. All rights reserved.
+
+// Script-generated file -- do not edit
+
+#pragma once
+
+#include "hipblas_arguments.hpp"
+
+template <typename T>
+inline inline hipblasStatus_t testing_asum(const Arguments& arg);
```
daineAMD, line 11: are these supposed to be `inline inline`?

TorreZuk: I think it should be "really inline"  :-) Very good find, more dead code.  That design didn't work as the template prototypes usage didn't work as compiler needed the implementation for the dispatch function pointers, will delete the file.

PR #536, opened 2022-10-14
Adding datatypes to type conversion function

daineAMD:


TorreZuk (approved):
Okay but what about some some logical order (increasing size?) to the enum sequence?

daineAMD:
I changed it to match the order in the enum declaration, looks to be the same order as in rocblas-types.h enum.

TorreZuk:
Okay that is fine but that order can't change to keep backward compatible, but I guess acts as reference.   The compiler should tell us if one is missed but it doesn't read well

PR #537, opened 2022-10-14
Cherry Pick Fix for old version rocm include/lib folders not removed on upgrade.

arvindcheru:
resolves #___

Summary of proposed changes:
- Cherry Pick Fix for old version rocm include/lib folders not removed on upgrade commit
-
-


TorreZuk (approved):
LGTM same merge approver considerations as rocblas

PR #539, opened 2022-11-01
Opt in for __half

daineAMD:
Adding use of __half for c++ users who define ~~HIPBLAS_USE_HIP_HALF~~ ROCM_MATHLIBS_API_USE_HIP_HALF.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,11 +117,6 @@ if( USE_CUDA )
     list( APPEND HIP_INCLUDE_DIRS "${HIP_ROOT_DIR}/include" )
 endif( )
 
-# HIP on Windows: xhip is required with clang++ to get __half defined
-if ( WIN32 )
-  add_compile_options(-xhip)
-endif()
-
 option(BUILD_CODE_COVERAGE "Build with code coverage enabled" OFF)
 if(BUILD_CODE_COVERAGE)
   add_compile_options(-fprofile-arcs -ftest-coverage)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,6 +117,11 @@ if( USE_CUDA )
     list( APPEND HIP_INCLUDE_DIRS "${HIP_ROOT_DIR}/include" )
 endif( )
 
+# HIP on Windows: xhip is required with clang++ to get __half defined
+if ( WIN32 )
+  add_compile_options(-xhip)
+endif()
+
 option(BUILD_CODE_COVERAGE "Build with code coverage enabled" OFF)
 if(BUILD_CODE_COVERAGE)
   add_compile_options(-fprofile-arcs -ftest-coverage)
--- a/clients/benchmarks/CMakeLists.txt
+++ b/clients/benchmarks/CMakeLists.txt
@@ -78,10 +78,9 @@ endif()
 
 target_link_libraries( hipblas-bench PRIVATE ${BLAS_LIBRARY} roc::hipblas Threads::Threads )
 
-# need mf16c flag for float->half convertion
-target_compile_options( hipblas-bench PRIVATE -mf16c ) # -Wno-deprecated-declarations )
+target_compile_options( hipblas-bench PRIVATE ) # -Wno-deprecated-declarations )
 
-target_compile_definitions( hipblas-bench PRIVATE HIPBLAS_BENCH HIPBLAS_BFLOAT16_CLASS ROCM_USE_FLOAT16 )
+target_compile_definitions( hipblas-bench PRIVATE HIPBLAS_BENCH HIPBLAS_BFLOAT16_CLASS HIPBLAS_USE_HIP_HALF )
 
 if( NOT USE_CUDA )
   target_link_libraries( hipblas-bench PRIVATE hip::host )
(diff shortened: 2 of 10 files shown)


TorreZuk (commented):
vs. ROCM_MATHLIBS_API_USE_HIP_HALF ?   Any other hip libraries to consider?
This is both internal and API so what you did also makes sense.

library/include/hipblas.h
```diff
@@ -84,7 +85,19 @@
...
 
 /*! \brief To specify the datatype to be unsigned short */
-typedef uint16_t hipblasHalf;
+
+#if __cplusplus < 201103L || !defined(HIPBLAS_USE_HIP_HALF)
+
+typedef struct hipblasHalf
+{
+    uint16_t data;
+} hipblasHalf;
```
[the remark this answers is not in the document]
daineAMD: You're right, this should be changed back.

daineAMD:
I've made the suggested changes, let me know what you think. I changed the define to ROCM_MATHLIBS_API_USE_HIP_HALF, I'm not sure if there are requests to make this change in other libraries but might as well be prepared.

TorreZuk:
> I've made the suggested changes, let me know what you think. I changed the define to ROCM_MATHLIBS_API_USE_HIP_HALF, I'm not sure if there are requests to make this change in other libraries but might as well be prepared.

Well it was just an option consider ....   Did you look at other hip math libs, any other __half users?  We should try to be more consistent across libs if possible.

clients/include/utility.h
```diff
@@ -120,9 +120,11 @@ inline bool hipblas_isnan(float arg)
 {
     return std::isnan(arg);
 }
+
 inline bool hipblas_isnan(hipblasHalf arg)
 {
-    return (~arg & 0x7c00) == 0 && (arg & 0x3ff) != 0;
+    auto half_data = static_cast<unsigned short>(arg);
+    return (~(half_data)&0x7c00) == 0 && (half_data & 0x3ff) != 0;
 }
```
amcamd, line 128: for consistency of formatting can you add spaces on either side of & to get (~(half_data) & 0x7c00) in place of (~(half_data)&0x7c00) . This is unless clang-format removes the spaces.

daineAMD:
It doesn't seem like this is a request for other libraries after looking around as far as I can tell. rocRAND, rocWMMA, rocPRIM/hipCUB/rocTHRUST all use __half and do not have an alternative. FFT and solver don't currently have any half-precision support. Looks like [rocPRIM](https://github.com/ROCmSoftwarePlatform/rocPRIM/blob/cd4397015c3f43d9625ce150d427acec20a35a8a/rocprim/include/rocprim/types.hpp#L161) used __half when available I think.

CMakeLists.txt
```diff
@@ -117,6 +117,11 @@ if( USE_CUDA )
     list( APPEND HIP_INCLUDE_DIRS "${HIP_ROOT_DIR}/include" )
 endif( )
 
+# HIP on Windows: xhip is required with clang++ to get __half defined
+if ( WIN32 )
+  add_compile_options(-xhip)
+endif()
```
TorreZuk, line 123: These blanket options are normally discouraged.   Can you set on targets if this is required?

library/include/hipblas.h
```diff
@@ -34,6 +34,7 @@
 
 #include "hipblas-export.h"
 #include "hipblas-version.h"
+#include <hip/hip_fp16.h>
```
TorreZuk, line 37: This include might be hidden within the enable use case #ifdef

daineAMD:
A couple of notes:
- It appears that many of the basic math operators for hip's __half are only __device__ calls, this was caught after trying it out myself on Windows. This is only an opt-in feature for now, so I've changed the PR to only test the __half type using a sample instead of incorporating this into an entire test suite as was done before.
- I've changed the option back to HIPBLAS_USE_HIP_HALF because as far as I can tell this won't be relevant to other libraries, so don't want users to be confused about this not working with other libraries.

clients/include/utility.h
daineAMD: Tried changing this and clang-format seems to remove them. clang-format also adds them in the later statement when I tried to remove them.

clients/benchmarks/CMakeLists.txt
```diff
@@ -78,7 +78,6 @@ endif()
 
 target_link_libraries( hipblas-bench PRIVATE ${BLAS_LIBRARY} roc::hipblas Threads::Threads )
 
-# need mf16c flag for float->half convertion
```
TorreZuk, line 81: I would keep this comment

clients/gtest/CMakeLists.txt
```diff
@@ -175,7 +175,6 @@ endif()
 
 target_link_libraries( hipblas-test PRIVATE ${BLAS_LIBRARY} roc::hipblas GTest::GTest Threads::Threads )
 
-# need mf16c flag for float->half convertion
```
TorreZuk, line 178: same here

TorreZuk (approved):
LG

clients/benchmarks/CMakeLists.txt
daineAMD: Ah yes I missed adding that back.

PR #544, opened 2022-11-08
Fixing geqrf return status for nullptrs, updating tests

daineAMD:
SWDEV-344553

- geqrf (and batched variants) will now return info = 0 when nullptrs are passed in with either n == 0 || m == 0. Note that batch_count == 0 will return bad info if a nullptr is passed in
- added some tests to tests this case and other bad input cases

TODO:
- the test suite is still not thorough, should add bad_args or something similar to rocBLAS as we aren't testing all cases for nullptrs, just added some ad-hoc testing for this fix
- will go through other rocSOLVER functions as this fix is probably needed in other places. Just wanted this to be ready quick/put in separately in case needed for a hotfix


jzuniga-amd (approved):
LGTM

PR #546, opened 2022-11-17
multiplot for relative performance

amcamd:
- code to review is in the following files:
  - benchmark_plot.sh
  - benchmark.py
  - plot.py 
- the code is similar to rocBLAS multiplot, but it adds the ability to plot performance on one platform relative to performance on another platform


TorreZuk (approved):
LGTM.  no images is fine but assume they look the same as rocblas

PR #547, opened 2022-11-17
merge staging de3eed into master on GO from CQE

daineAMD:


TorreZuk (approved):
LGTM the manifest is going to commit de3eed5 so not sure if that last merge commit matters I see in this PR.

PR #549, opened 2022-11-23
Increase gemm_ex tolerance for gfx11

daineAMD:
SWDEV-369347

Increasing tolerance for gemm_ex gfx11 tests, see rocmsoftwareplatform/rocblas-internal/pull/1518. Uses a much lower tolerance than rocBLAS does currently as it doesn't seem necessary in hipBLAS tests. This may cause failures in the future, but could also catch things that rocBLAS missed.


clients/include/near.h
```diff
@@ -87,4 +87,39 @@ HIPBLAS_CLANG_STATIC constexpr double error_tolerance = 0.0;
...
+    = 1 / 10000.0;
+
+template <>
+HIPBLAS_CLANG_STATIC constexpr double sum_error_tolerance_for_gfx11<float, hipblasHalf, hipblasHalf>
+    = 1 / 10000.0;
+
+template <>
+HIPBLAS_CLANG_STATIC constexpr double
+    sum_error_tolerance_for_gfx11<hipblasHalf, hipblasHalf, hipblasHalf>
+    = 1 / 10000.0;
```
TorreZuk, line 113: Your build may not be using the new assembly instructions yet, it is in progress.  I would test with navi32 to see the future, or could go with what rocblas uses.  Maybe your tests are all small enough K ?

clients/include/testing_gemm_ex.hpp
```diff
@@ -225,8 +226,21 @@ inline hipblasStatus_t testing_gemm_ex_template(const Arguments& arg)
 
             if(unit_check)
             {
-                unit_check_general<Tc>(M, N, ldc, hC_gold, hC_host);
-                unit_check_general<Tc>(M, N, ldc, hC_gold, hC_device);
+                // check for mixed precision with 16 bit input and 32 bit computation
```
TorreZuk, line 229: Your comment implies 32 bit computation, is Tex computation or output, as half == Tex == Ta is included.

TorreZuk (approved):
LGTM but I would do the test on navi32 or increase the tolerance so we don't get another ticket on the next promotion.

clients/include/near.h
daineAMD: I built in the docker container in the ticket and it failed using unit_check() and passed with these tolerances. I don't mind changing it to match rocBLAS if we're concerned with future changes affecting this, though, just seemed excessive to use 1/10.0 when 1/10000.0 seemed to work.

clients/include/testing_gemm_ex.hpp
amcamd: Tex should be computation, it defaults to output (Tc) if not specified. Below from lines 43 and 44 is the prototype for testing_gemm_ex_template:

template <typename Ta, typename Tb = Ta, typename Tc = Tb, typename Tex = Tc>
inline hipblasStatus_t testing_gemm_ex_template(const Arguments& arg)

amcamd (approved):

clients/include/near.h
amcamd: If it passes in the docker I would favor your tighter tolerance. You could add a comment that if k is large the tolerance would need to increase.

clients/include/testing_gemm_ex.hpp
daineAMD: Both hgemm (with half input/output/computation) and gemm_ex with half input/output and float computation are failing, thus the `if()` statement including both. The comment is outdated and I've updated it in the next commit.

clients/include/near.h
TorreZuk: Okay K is so small it isn't really doing any reduction, but still would test on gfx1101

daineAMD: Looking at the tests, it looks like the largest k is 1027, (M=925, N=1026, K=1027).
Tests all passed on gfx1101 just now.

PR #551, opened 2022-11-29
match rocblas helpers

TorreZuk:
* Only saw the python for the yaml compiling, pyyaml , Daine weren't the others just from rocblas (only for tensile build not runtime)


rdeps.xml
```diff
@@ -9,10 +9,7 @@
   </os>
   <os names="all">
     <pip>
-      <pkg name="PyYAML">pyyaml</pkg>
-      <pkg name="six">six</pkg>
-      <pkg name="wheel">wheel</pkg>
-      <pkg name="virtualenv">virtualenv</pkg>
```
daineAMD, line 15: Yes I think that all of these are only needed for venv/Tensile, should be fine to remove.

daineAMD (approved):

PR #554, opened 2022-12-02
Allowing spaces in windows hip directories

daineAMD:


TorreZuk (approved):
Assume you tested with "hip SDK" in the HIP_PATH env as CI won't test this.
LGTM

PR #555, opened 2022-12-03
rmake for linux, rmake cleanup

daineAMD:
Need to go through and test all of the options manually yet.
Aside: Will enable blis cpu reference on linux in another PR next week, using openblas for windows.


clients/CMakeLists.txt
```diff
@@ -77,13 +77,16 @@ if( BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
   if( NOT WIN32 )
     # Linking lapack library requires fortran flags
     find_package( cblas REQUIRED CONFIG )
-    if (LINK_BLIS)
-      set( BLIS_INCLUDE_DIR ${BUILD_DIR}/deps/blis/include/blis )
```
TorreZuk, line 81: Not sure what you are doing here you are in a NOT WIN32 already.   Are you dropping blis?

daineAMD: It doesn't appear that we're using blis at all. I changed this code back to what it was, but we aren't defining LINK_BLIS in install.sh or rmake.py so it isn't used anyway. I'll work on adding blis in another PR.

TorreZuk: So what is the goal of this PR, rmake.py should work find on linux?  Yes blis or other reference another PR.  OpenBLAS on windows covers rocblas (solver have to check?), on linux should get solver input as well.

daineAMD: Yes, rmake.py now works on linux after this PR.

rmake.py
```diff
@@ -190,22 +213,24 @@ def config_cmd():
     if args.static_lib:
         cmake_options.append( f"-DBUILD_SHARED_LIBS=OFF" )
 
-    if args.skip_ld_conf_entry:
+    if args.relocatable:
+        rocm_rpath = os.getenv( 'ROCM_RPATH', "/opt/rocm/lib:/opt/rocm/lib64")
+        cmake_options.append( f'-DCMAKE_SHARED_LINKER_FLAGS=" -Wl,--enable-new-dtags -Wl,--rpath,{rocm_rpath}"' )
+
+    if args.skip_ld_conf_entry or args.relocatable:
         cmake_options.append( f"-DROCM_DISABLE_LDCONFIG=ON" )
```
TorreZuk, line 221: This may need to change once the PR goes in changing install.sh but matching fine for now

TorreZuk (approved):
LGTM if all passing

PR #556, opened 2022-12-08
Changing Documentation Template

SergiyKostrov:
Changing Documentation Template - hipBLAS


saadrahim:
@SergiyKostrov a few settings in conf.py were not carried forwards relating to the theme. Can you update that? If you look at the doc build, the theme is still the standard blue RTD them. We are switching to the sphinx book theme.

saadrahim:
Also, I think you may have copied settings from the rocBLAS devel branch. Not the PR 1560.

PR #571, opened 2023-03-14
Static hipblas_fortran_client library

daineAMD:
Should fix the problem of finding libhipblas_fortran_client.so from hipblas-clients package. Making hipblas_fortran_client a static library, the same as rocblas_fortran_client.
@lawruble13 can you take a look? I see you make the original [rocBLAS changes](https://github.com/rocmsoftwareplatform/rocblas-internal/pull/1099) and [hipBLAS changes](https://github.com/rocmsoftwareplatform/hipblas/pull/463).


TorreZuk (approved):
LGTM

lawruble13:
This looks good to me, although I would note that with the common fortran library compiled statically there's no longer a reason for the clients-common package, so that could be safely removed.

lawruble13 (approved):

PR #573, opened 2023-03-16
Removing support for packed int8

daineAMD:
Removes support for:
- `hipblasInt8Datatype_t` enum
- `hipblasSetInt8Datatype`
- `hipblasGetInt8Datatype`
- related internal functions

Removes testing, and removes some documentation. Will add in CHANGELOG at FC unless there's somewhere better to include this (deprecation.rst?)

### DIFF
--- a/clients/common/utility.cpp
+++ b/clients/common/utility.cpp
@@ -1,5 +1,5 @@
 /* ************************************************************************
- * Copyright (C) 2016-2022 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (C) 2016-2023 Advanced Micro Devices, Inc. All rights reserved.
  *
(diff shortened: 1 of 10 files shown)


TorreZuk (approved):
Great to see the packed int go away.   Aren't there tests that need to be removed?
LGTM if all tests pass

daineAMD:
> Great to see the packed int go away. Aren't there tests that need to be removed? LGTM if all tests pass

I think the only real "test" was what was in [testing_gemm_ex.hpp](https://github.com/ROCmSoftwarePlatform/hipBLAS/pull/573/files#diff-909e3e533c39222eb952ca6156fce0ca031821639e5887f9205d53c803463898L108)

PR #577, opened 2023-03-21
Trmm deprecation with HIPBLAS_V1

amcamd:
- contains deprecation code and sample. 
- Does not contain tests with hipblas-test


library/src/amd_detail/hipblas.cpp
```diff
@@ -13992,6 +13993,489 @@ catch(...)
...
+                                                             hipDiagonalToHCCDiagonal(diag),
+                                                             m,
+                                                             n,
+                                                             alpha,
+                                                             A,
+                                                             lda,
+                                                             B,
+                                                             ldb,
+                                                             C,
+                                                             ldc));
```
TorreZuk, line 14026: You don't want to define ROCBLAS_V3 and test that system here and use the new name?

TorreZuk (approved):
It looks like it matches the pattern you set in rocBLAS.  Good to get an additional review and I will review the build log, but expect deprecation warnings only from a sample, or none if trmm is only usage and built using -DHIPBLAS_V1.  Tests in separate PR?

clients/samples/example_strmm.cpp
```diff
@@ -0,0 +1,264 @@
...
+                __LINE__);                        \
+        exit(EXIT_FAILURE);                       \
+    }
+#endif
+
+#ifndef CHECK_HIPBLAS_ERROR
+#define CHECK_HIPBLAS_ERROR(error)                              \
+    if(error != HIPBLAS_STATUS_SUCCESS)                         \
+    {                                                           \
+        fprintf(stderr, "rocBLAS error: ");                     \
```
daineAMD, line 50: Should probably be "hipBLAS error" but I'm guessing this is in all/most of our samples so fine to ignore.

daineAMD (approved):

clients/samples/example_strmm.cpp
amcamd: Will update in another PR.

library/src/amd_detail/hipblas.cpp
amcamd: I think it is works as it is, and it seems to me to be more similar in amd_detail and nvidia_detail. An alternative would be to define ROCBLAS_V3 and use the new name. Below is how it works as it is: 

The librocblas.so contains both rocblas_strmm(... A, lda, B, ldb...) and rocblas_strmm_outofplace(...A, lda, B, ldb, C, ldc ...). 

Same for libhipblas.so, it contains both hipblasStrmm( ... A, lda, B, ldb, ...) and hipblasStrmmOutofplace(... A, lda, B, ldb, C, ldc, ...). 

If you compile without HIPBLAS_V1 you can call either hipblasStrmm or hipblasStrmmOutofplace . You get hipBLAS deprecation messages for both hipblasStrmm and hipblasStrmmOutofplace (rocBLAS deprecation messages are turned off at line 23 with # define ROCBLAS_NO_DEPRECATION_WARNINGS)

If you compile with HIPBLAS_V1 then any call you make to hipblasStrmm(  A, lda, B, ldb, C, ldc, ...) in your code is converted to a call to hipblasStrmmOutofplace(..A, lda, B, ldb, C, ldc)  by cpp. Calls to hipblasStrmm(.. A, lda, B, ldb ...) will give link errors because they are converted to hipblasStrmmOutofplace( ... A, lda, B, ldb, C, ldc ...) and there is an argument mismatch.

PR #579, opened 2023-03-27
don't copy extra dlls

TorreZuk:
* avoids copying in extra rocm math libs which may contaminate local build


TorreZuk (commented):
Passed my local win32 CI

daineAMD (approved):

TorreZuk:
Went by my private CI so if any trouble get me to hotfix, not sure why I don't see windows CI running (I'll watch for it though)

PR #585, opened 2023-04-06
Add rocSPARSE dependency to math-ci

cgmb:
The rocSOLVER library has added a dependency on rocSPARSE, so it may be required to successfully link with rocSOLVER.

### DIFF
--- a/.jenkins/multicompiler.groovy
+++ b/.jenkins/multicompiler.groovy
@@ -17,7 +17,7 @@ def runCI =
 
     //customize for project
     prj.paths.build_command = buildCommand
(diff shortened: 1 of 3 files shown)


daineAMD (approved):

cgmb:
> Are the solver functions that need sparse being exposed in hipBLAS?

I don't think so.

> If not could an on demand dependency load be used in solver? More of a solver question than on this PR.

That might be something to consider. There is a `BUILD_WITH_SPARSE` CMake option for rocSOLVER, but the CI currently provides hipBLAS with a copy of rocSOLVER where that was set to `ON`. If the CI used a copy of rocSOLVER built with that option set to `OFF`, then hipBLAS would not need to have rocSPARSE installed.

PR #595, opened 2023-05-19
Updating changelog for rocm 5.7

daineAMD:
Looks like not really any changes for hipBLAS this release.

https://github.com/rocmsoftwareplatform/hipblas/compare/release/rocm-rel-5.6...develop

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 ### Changed
 - updated documentation requirements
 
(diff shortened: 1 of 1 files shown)


TorreZuk (approved):
LGTM but even though a transitive dependency change you could point out rocSOLVER which is a dependency now depends on rocSPARSE?   Or was that in previous release....

PR #598, opened 2023-05-24
Remove old docs action

samjwu:
The old docs workflow of uploading a zip upon push to the develop or master branch is replaced by automatic Read the Docs (.com, .org is the community version with less features to be phased out) builds on each PR


TorreZuk (approved):
Thanks for the quick response.

Issue #603, opened 2023-05-30
Building rocBLAS for CUDA backend fails


torrance:
I'm building using CMake directly[1] for CUDA backend:

```
# On branch release/rocm-rel-5.5
mkdir build && cd build
cmake -DUSE_CUDA=ON -DHIP_ROOT_DIR=/opt/rocm ..
make

Scanning dependencies of target hipblas_fortran
[ 20%] Building Fortran object library/src/CMakeFiles/hipblas_fortran.dir/hipblas_module.f90.o
[ 40%] Linking Fortran shared library libhipblas_fortran.so
[ 40%] Built target hipblas_fortran
[ 60%] Building CXX object library/src/CMakeFiles/hipblas.dir/nvidia_detail/hipblas.cpp.o
In file included from /home/torrance/hipBLAS/library/src/nvidia_detail/hipblas.cpp:27:
/usr/local/cuda/include/cublas_v2.h:59:2: error: #error "It is an error to include both cublas.h and cublas_v2.h"
   59 | #error "It is an error to include both cublas.h and cublas_v2.h"
      |  ^~~~~
```

I can compile with the following changes, though I don't know if this is properly functional or not:

```diff
diff --git a/library/src/nvidia_detail/hipblas.cpp b/library/src/nvidia_detail/hipblas.cpp
index b60a31d..08a6257 100644
--- a/library/src/nvidia_detail/hipblas.cpp
+++ b/library/src/nvidia_detail/hipblas.cpp
@@ -23,7 +23,7 @@

 #include "hipblas.h"
 #include "exceptions.hpp"
-#include <cublas.h>
+//#include <cublas.h>
 #include <cublas_v2.h>
 #include <cuda_runtime_api.h>
 #include <hip/hip_runtime.h>
@@ -350,7 +350,7 @@ hipblasStatus_t hipblasGetPointerMode(hipblasHandle_t handle, hipblasPointerMode
 try
 {
     cublasPointerMode_t cublasMode;
-    cublasStatus        status = cublasGetPointerMode((cublasHandle_t)handle, &cublasMode);
+    cublasStatus_t        status = cublasGetPointerMode((cublasHandle_t)handle, &cublasMode);
     *mode                      = CudaPointerModeToHIPPointerMode(cublasMode);
     return hipCUBLASStatusToHIPStatus(status);
 }
```

[...]

PR #605, opened 2023-05-30
Fix for cuda 12

daineAMD:
See #603


TorreZuk (approved):
LGTM. So we only use v2 headers.   I guess you will hotfix PR this to 5.7?

PR #610, opened 2023-06-04
Fix spelling errors

cgmb:
Fixes spelling errors noticed while packaging hipblas for Debian.


amcamd (approved):
It is good to have these spelling errors corrected.

PR #611, opened 2023-06-04
Use new-style includes in samples

cgmb:
Users of hipBLAS should be including the header as <hipblas/hipblas.h>, so it is nice to use that style of include in the samples.


amcamd (approved):
LGTM

PR #619, opened 2023-06-22
Hotfix for asan build with flang

daineAMD:
SWDEV-407304

Basically we shouldn't have a module split between two .f90 files. Disclaimer: I'm in no way a fortran expert.

Changes:
- add gfortran dep to clients for lapack when built with flang
- rename `hipblas_fortran.f90` -> `hipblas_fortran_blas.f90`. Contains BLAS functions (for clients), doesn't have hipblas_interface module anymore
- remove hipblas_interface module from `hipblas_fortran_solver.f90`
- create `hipblas_fortran_no_solver_module.f90` which includes `hipblas_fortran_blas.f90` and has `hipblas_interface` module. Only built with --nosolver. This could be dealt with a preprocessor macro in `hipblas_fortran_module.f90` instead but not exactly sure on best practices/compiler support.
- create `hipblas_fortran_module.f90` which includes `hipblas_fortran_blas.f90`, `hipblas_fortran_solver.f90` and has `hipblas_interface` module.

We should actually add support for different fortran compilers, this is just ad-hoc and seems to work, but not sure of the future of fortran here.


TorreZuk (approved):
LGTM to separate the overlap this way.
So flang will be used by them but we'll linking gfortran is okay?   Did someone change our build to use flang?  CI is failing for other reasons so can't see.

daineAMD:
I tested locally by changing the compiler to flang and it seemed to work. The linking is just needed for the reference library, not really sure.
I'm not clear on how they're building hipBLAS in the ticket, they have their own bash script I guess. I'll wait to see if it fixes their problem using their build system (as well as CI with gfortran) before merging this.

clients/benchmarks/CMakeLists.txt
```diff
@@ -73,7 +73,7 @@ target_include_directories( hipblas-bench
 )
 
 if (NOT WIN32)
-    target_link_libraries( hipblas-bench PRIVATE hipblas_fortran_client lapack cblas stdc++fs )
+    target_link_libraries( hipblas-bench PRIVATE hipblas_fortran_client lapack cblas stdc++fs gfortran )
```
cgmb, line 76: I can't imagine that linking libgfortran is the way to fix flang.

cgmb (changes requested):

clients/benchmarks/CMakeLists.txt
cgmb: This change can be dropped. I've prepared updates to the build infrastructure so that all Fortran components are built with the same toolchain. In general, that is the only way to ensure that all the proper runtime libraries are linked. If you start explicitly linking runtimes like in this patch, you're embedding assumptions about the toolchain into your build.

clients/gtest/CMakeLists.txt
```diff
@@ -161,7 +161,7 @@ target_include_directories( hipblas-test
 )
 
 if (NOT WIN32)
-    target_link_libraries( hipblas-test PRIVATE hipblas_fortran_client lapack cblas stdc++fs )
+    target_link_libraries( hipblas-test PRIVATE hipblas_fortran_client lapack cblas stdc++fs gfortran )
```
cgmb: This can be dropped as well.

clients/include/hipblas_fortran_module.f90
```diff
@@ -0,0 +1,32 @@
...
+
+module hipblas_interface
+    use iso_c_binding
+    use hipblas
+
+contains
+    include 'hipblas_fortran_blas.f90'
+    include 'hipblas_fortran_solver.f90'
+
+end module hipblas_interface
```
cgmb: [POSIX 2018 §3.206](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206) defines a line in a text file as

> A sequence of zero or more non- <newline> characters plus a terminating <newline> character.

I would recommend ensuring that all source files end in a newline. Most text processing tools these days are robust to this sort of thing, but at the very least it makes the output of `cat` a bit nicer.

clients/include/hipblas_fortran_no_solver_module.f90
```diff
@@ -0,0 +1,31 @@
...
+!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+module hipblas_interface
+    use iso_c_binding
+    use hipblas
+
+contains
+    include 'hipblas_fortran_blas.f90'
+
+end module hipblas_interface
```
cgmb: This should also end in a newline.

cgmb (changes requested):
I've prepared updates that fix flang in the build infrastructure. That problem was in their configuration, not in the hipBLAS cmake code, so I recommend against adding workarounds here. (i.e., "add gfortran dep to clients for lapack when built with flang" is unnecessary.) However, the splitting of the interface to fix the no_solver build is still worth doing.

cgmb (approved):

PR #621, opened 2023-06-27
Deprecating hipblasDatatype_t for hipDataType; adding HIPBLAS_V2 Compile Option

daineAMD:
This PR deprecates the use of hipblasDatatype_t to be replaced with hipDataType in a future release.

Library
---------
- gives deprecation warnings for hipblasDatatype_t
- adds option to compile with HIPBLAS_V2 which:
    - typedefs hipblasDatatype_t to hipDataType. All functions which use this type will thus have new interfaces using hipDataType.
- when NOT compiling with HIPBLAS_V2:
    - all functions using hipblasDatatype_t will stay the same.
    - in this PR, I've introduced hipblasScalEx_v2 which can take hipDataType instead of hipblasDatatype_t. This is only here for ease of transition and will be replaced by hipblasScalEx once hipblasDatatype_t is removed. I've only done this for scalEx for now, but if approved will go through for all functions using hipblasDatatype_t

Clients
---------
- adds sample program to show use of hipblasScalEx with hipDataTypes when compiling with HIPBLAS_V2
- adds hipblas_v2-test and hipblas_v2-bench which are compiled using HIPBLAS_V2 option, thus can use new hipblasScalEx interface with hipDataType
- adds testing of hipblas_v2-test to jenkins, although I don't know what I'm doing here so hopefully someone with experience can double check/can wait for CI confirmation that this works
- until other functions are updated hipblas_v2-test and hipblas_v2-bench will not work with functions using hipblasDatatype_t, e.g. hipblasNrm2Ex

Note that the interface will change _further_ in a future PR where `hipblasComputeType_t` will be added and used in place of `hipblasDatatype_t`/`hipDataType` where applicable. This will be done in the same release cycle.

Note: in bf5fe982 I've moved the #include of cuda into hipblas.h for cuda compilation. It seems that hip's library_types.h doesn't include cuda's library_types.h correctly, but cublas does. Testing this out on CI for now.


clients/gtest/CMakeLists.txt
```diff
@@ -215,9 +240,11 @@ if (WIN32)
   )
   foreach( file_i ${third_party_dlls})
     add_custom_command( TARGET hipblas-test POST_BUILD COMMAND ${CMAKE_COMMAND} ARGS -E copy ${file_i} ${PROJECT_BINARY_DIR}/staging/ )
+    add_custom_command( TARGET hipblas_v2-test POST_BUILD COMMAND ${CMAKE_COMMAND} ARGS -E copy ${file_i} ${PROJECT_BINARY_DIR}/staging/ )
```
TorreZuk, line 243: Unless you expect only one target to be built you don't need to copy the same files twice.

clients/gtest/CMakeLists.txt
```diff
@@ -215,9 +240,11 @@ if (WIN32)
...
   endforeach( file_i )
 
   add_custom_command( TARGET hipblas-test POST_BUILD COMMAND ${CMAKE_COMMAND} ARGS -E copy_directory ${ROCBLAS_PATH}/bin/rocblas/library/ ${PROJECT_BINARY_DIR}/staging/library/)
+  add_custom_command( TARGET hipblas_v2-test POST_BUILD COMMAND ${CMAKE_COMMAND} ARGS -E copy_directory ${ROCBLAS_PATH}/bin/rocblas/library/ ${PROJECT_BINARY_DIR}/staging/library/)
```
TorreZuk, line 247: Same here

clients/include/hipblas_common.yaml
```diff
@@ -3,24 +3,39 @@
...
-        i32_c: 166
-        u32_c: 167
-        bf16_r: 168
-        bf16_c: 169
+        f32_r:   0
+        f64_r:   1
+        f16_r:   2
+        i8_r:    3
+        f32_c:   4
+        f64_c:   5
```
TorreZuk, line 16: I guess this is the hipDataType enum, mayb a comment to say such if true

library/include/hipblas.h
```diff
@@ -434,7 +435,34 @@ typedef enum
...
+#define HIPBLAS_C_32I HIP_C_32I
+#define HIPBLAS_C_32U HIP_C_32U
+#define HIPBLAS_R_16B HIP_R_16BF
+#define HIPBLAS_C_16B HIP_C_16BF
+#define HIPBLAS_DATATYPE_INVALID hipDataType(255)
+
+#else
+
+// clang-format off
+HIPBLAS_DEPRECATED_MSG("hipblasDatatype_t is deprecated, and it will be replaced completely by hipDataType in the future. Compiling with -DHIPBLAS_V2 will replace hipblasDatatype_t in the API with hipDataType.")
```
TorreZuk, line 464: I expect this gets reported a large many times as on the enum right? Maybe can make a bit more concise?

library/include/hipblas.h
```diff
@@ -19873,6 +19945,9 @@ HIPBLAS_EXPORT hipblasStatus_t hipblasScalEx(hipblasHandle_t   handle,
 
     - Supported types are determined by the backend. See rocBLAS/cuBLAS documentation.
 
+    When compiling with HIPBLAS_V2, hipblasScalBatchedEx accepts hipDataType for alphaType,
```
TorreZuk, line 19948: Is there a different include file hipblas_v2.h as well to force this?   or just -DHIPBLAS_V2 ?  Could say "HIPBLAS_V2 define" or similar.  I presume it matches rocblas methodology, so fine.

library/src/nvidia_detail/hipblas.cpp
```diff
@@ -12582,6 +12633,30 @@ catch(...)
...
+                                                   HIPDatatypeToCudaDatatype_v2(alphaType),
+                                                   x,
+                                                   HIPDatatypeToCudaDatatype_v2(xType),
+                                                   incx,
+                                                   HIPDatatypeToCudaDatatype_v2(executionType)));
+}
+catch(...)
+{
+    return exception_to_hipblas_status();
+}
```
TorreZuk, line 12658: I guess you don't want to try my macro ILP64 strategy here :-)   I guess not if you can delete one soon.

TorreZuk (approved):
Pattern LGTM but Andrew can review better for this.

library/include/hipblas.h
daineAMD: Yeah when building hipBLAS it really makes it difficult to see compilation. I shortened it, but best to use HIPBLAS_NO_DEPRECATED_WARNINGS for devs.

library/include/hipblas.h
daineAMD: Just HIPBLAS_V2, no hipblas_v2.h. Updated the comment a bit.

TorreZuk:
Are you defining HIPBLAS_NO_DEPRECATED_WARNINGS for the library and test and benchmark as compilation defines?

TorreZuk:
Another weak github visual design: see my comment on your last commit bd3aabb

library/include/hipblas.h
```diff
@@ -19864,6 +19901,45 @@ HIPBLAS_EXPORT hipblasStatus_t hipblasScalEx(hipblasHandle_t   handle,
...
+    @param[in]
+    incx      [int]
+              specifies the increment for the elements of x.
+    @param[in]
+    executionType [hipDataType]
+                   specifies the datatype of computation.
+
+    ********************************************************************/
+HIPBLAS_EXPORT hipblasStatus_t hipblasScalEx_v2(hipblasHandle_t handle,
+                                                int             n,
```
amcamd, line 19935: I do not think we need to document the `hipblasScalEx_v2` function. We hope the user compiles with `-D HIPBLAS_V2` and calls hipblasScalEx and hipblasScalEx is converted to hipblasScalEx_v2 by the cpp preprocessor. 
Another way it could be documented is like https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/release/rocm-rel-5.6/library/include/internal/rocblas-functions.h#L12763

library/src/nvidia_detail/hipblas.cpp
```diff
@@ -226,6 +226,57 @@ cudaDataType_t HIPDatatypeToCudaDatatype(hipblasDatatype_t type)
     }
 }
 
+cudaDataType_t HIPDatatypeToCudaDatatype_v2(hipDataType type)
+{
```
amcamd, line 230: Can this be `HIPDatatypeToCudaDatatype` rather than `HIPDatatypeToCudaDatatype_v2` ?

amcamd (commented):
Preliminary review, will add additional comments

library/src/nvidia_detail/hipblas.cpp
daineAMD: No, the original, which takes a hipblasDatatype_t, is called (mistakenly?) `HIPDatatypeToCudaDatatype` so can't overload it.

PR #626, opened 2023-06-30
hipDataType for axpyEx, dotEx, nrm2Ex, rotEx, and trsmEx

daineAMD:
Follow up to #621
Adds functions accepting hipDataType for axpyEx, dot(c)Ex, nrm2Ex, rotEx, and trsmEx. The only one remaining is gemmEx, but I'll include that in another PR which introduces hipblasComputeType.


library/include/hipblas.h
```diff
@@ -19276,16 +19676,20 @@ HIPBLAS_EXPORT hipblasStatus_t hipblasAxpyStridedBatchedEx(hipblasHandle_t   han
...
     incx      [int]
               specifies the increment for the elements of y.
     @param[in]
     y         device pointer storing vector y.
     @param[in]
-    yType [hipblasDatatype_t]
-          specifies the datatype of vector y.
+    yType     [hipblasDatatype_t] [DEPRECATED]
+              specifies the datatype of vector y.\n
+    yType     [hipDataType]
```
TorreZuk: Newlines \n showing up.

TorreZuk (approved):
LGTM, not sure if I suggested already but hipblas_v2.h could just
#define HIPBLAS_V2 1
#include "hiplas.h"
for those who dont want to touch Makefiles.

library/include/hipblas.h
daineAMD: For doxygen formatting, should create a newline in the docs. I changed it slightly in the next commit to look better. Note that if you're looking at CI docs, the API is erroneously under "guidelines", I'll make a ticket for it.

TorreZuk: Okay I found it under guidelines and it looks okay.   Maybe in that ticket you can push solver API after blas extensions.  Also will the hipblasDataType enum be deprecated in the docs in that PR or this one?

daineAMD: Sure, will do.
Should have been in this or the last PR, but I have another one coming with changes to gemmEx so I'll just put it in that one.

PR #629, opened 2023-07-06
Deprecation for hipblasGemmEx

daineAMD:
This deprecates hipblasGemmEx to use `hipDataType` instead of `hipblasDatatype_t`, along with introducing and using `hipblasComputeType_t` instead of `hipblasDatatype_t` for compute_type for hipblasGemmEx.

rocBLAS support is specified in library/include/hipblas.h. I've opted to not include support for _PEDANTIC types, even though we technically could support fp16, fp32, and fp64 pedantic (via hgemm, sgemm, and dgemm). To include this support would be a commitment from rocBLAS/Tensile so I've left it for now and we can discuss if we want to include later.


TorreZuk (approved):
Didn't spot anything.  Can you you document in the ticket what the test coverage plan is.  V2 appears less for gemm_ex but when the old API version is dropped will it get all the original test coverage?   Probably also doc what compute types are not tested as well.

daineAMD:
All rocBLAS supported compute types are tested, I guess we're lacking coverage for cuBLAS backend but I don't think we have to be thorough there, I'll add a comment in the test code and create/update ticket, along with test coverage plan.
Yes, once old API is dropped I'll remove the added _V2 tests and point all of the previous test cases (along with new ones with further compute coverage) to the new API.

library/include/hipblas.h
```diff
@@ -18426,6 +18543,58 @@ HIPBLAS_EXPORT hipblasStatus_t hipblasGemmEx(hipblasHandle_t    handle,
...
+                                                 const void*          beta,
+                                                 void*                C[],
+                                                 hipDataType          cType,
+                                                 int                  ldc,
+                                                 int                  batchCount,
+                                                 hipblasComputeType_t computeType,
+                                                 hipblasGemmAlgo_t    algo)
+
+        #else // Deprecated
+
```
amcamd: to be consistent can this be changed from `Deprecated` to `[DEPRECATED]`

library/include/hipblas.h
```diff
@@ -18534,6 +18735,64 @@ HIPBLAS_EXPORT hipblasStatus_t hipblasGemmBatchedEx(hipblasHandle_t    handle,
...
+                                                        void*                C,
+                                                        hipDataType          cType,
+                                                        int                  ldc,
+                                                        hipblasStride        strideC,
+                                                        int                  batchCount,
+                                                        hipblasComputeType_t computeType,
+                                                        hipblasGemmAlgo_t    algo)
+
+        #else // Deprecated
+
```
amcamd: for consistency `Deprecated` -> `[DEPRECATED]`

clients/samples/CMakeLists.txt
```diff
@@ -54,17 +55,19 @@ if( NOT TARGET hipblas )
   endif( )
 endif( )
 
-list (APPEND hipblas-example-executables hipblas-example-sscal hipblas-example-scal-ex-v2 hipblas-example-strmm_v1 hipblas-example-sgemm hipblas-example-sgemm-strided-batched hipblas-example-c hipblas-example-hip-complex-her2 hipblas-example-hgemm-half)
+list (APPEND hipblas-example-executables hipblas-example-sscal hipblas-example-scal-ex-v2 hipblas-example-strmm_v1 hipblas-example-sgemm hipblas-example-sgemm-strided-batched hipblas-example-gemmex_v2 hipblas-example-c hipblas-example-hip-complex-her2 hipblas-example-hgemm-half)
 if( CMAKE_CXX_COMPILER MATCHES ".*/hipcc$" )
```
amcamd, line 57: Maybe make the change  `hipblas-example-gemmex_v2` -> `hipblas-example-gemm_ex_v2` or `hipblas-example-gemmEx_v2`

clients/include/testing_gemm_ex.hpp
```diff
@@ -132,18 +133,22 @@ inline hipblasStatus_t testing_gemm_ex_template(const Arguments& arg)
...
                                             lda,
                                             dB,
                                             b_type,
                                             ldb,
-                                            &h_beta_Tc,
+                                            &h_beta_Tex,
                                             dC,
                                             c_type,
                                             ldc,
+#ifdef HIPBLAS_V2
```
amcamd, line 147: should a_type, b_type, c_type be hipDatatype when HIPBLAS_V2 is defined?

amcamd (approved):

clients/include/testing_gemm_ex.hpp
daineAMD: Yes, but in hipblas.h I have:
```
#ifdef HIPBLAS_V2
typedef hipDataType hipblasDatatype_t;
...
#endif
```
To make the switch easier.

library/include/hipblas.h
daineAMD: Updated in next commit, along with renaming the sample executable to hipblas-example-gemmEx_v2.

PR #631, opened 2023-07-12
Fixing --yaml option for hipblas-test and hipblas_v2-test

daineAMD:
- Small hacky solution to getting hipblas-test and hipblas_v2-test to both work with hipblas_common.yaml and --yaml option
- Remove duplicate hipblas_common.yaml in scripts directory
- Fix to hipblas_common.yaml for --inplace argument


TorreZuk (approved):
LGTM  if you could run smoke test with and without V2?

daineAMD:
> LGTM if you could run smoke test with and without V2?

Yep, both `./hipblas-test --yaml hipblas_smoke.yaml` and `./hipblas_v2-test --yaml hipblas_smoke.yaml` pass all tests.

PR #633, opened 2023-07-14
Some hipBLAS documentation updates.

daineAMD:
There's plenty of work to do to improve hipBLAS documentation. Would be good to have it as good as the rocBLAS documentation since it's likely to be used by more general users. This is just an incremental improvement prompted by [LWPMLSE-463](https://ontrack-internal.amd.com/browse/LWPMLSE-463). Will add documentation regarding atomics after this goes in.


docs/functions.rst
```diff
@@ -2,13 +2,31 @@
...
+                float *y, int incy );
+
 Naming conventions
 ==================
 
-hipBLAS follows the following naming conventions,
+hipBLAS follows the following naming conventions:
 
 - Big case for matrix, e.g. matrix A, B, C   GEMM (C = A*B)
 - Lower case for vector, e.g. vector x, y    GEMV (y = A*x)
```
TorreZuk: Upper case ...

TorreZuk (approved):
Good start.  If matches general pattern of rocblas may be simpler to maintain.

PR #636, opened 2023-07-18
Allow hipBLAS enums to be defined by hipSOLVER

tfalders:
This is a companion PR to https://github.com/ROCmSoftwarePlatform/hipSOLVER/pull/180.

Currently, there are difficulties to hipify cuSOLVER code to hipSOLVER, because cuSOLVER uses enums from cuBLAS but hipSOLVER does not use enums from hipBLAS (we use hipsolverOperation_t instead of hipblasOperation_t, for example). We wish to fix this in the next release and have hipSOLVER use hipBLAS enums.

We identified 3 potential solutions to this problem:
1. Add hipBLAS as an optional dependency to hipSOLVER. This approach turned out to be quite complicated and doesn't appear to have the necessary support in rocm-cmake.
2. Add hipBLAS as a hard dependency to hipSOLVER. This approach seems a little excessive since we're only borrowing a few enums from hipBLAS.
3. Define hipBLAS enums in hipSOLVER, using macros to prevent name conflicts in the headers and static asserts to ensure the values are always as expected. This will continue to allow hipSOLVER to be used independently of hipBLAS and is the approach taken in this PR.

I've run some tests to ensure that hipSOLVER and hipBLAS can be used together with these changes, and although I had some initial test errors, after purging my build directory and doing a clean rebuild, those errors appear to have gone away.


TorreZuk (commented):
As strange as this PR is,  I don't have a particular problem with it.
If going this route I would say best to extract this code into hipblas-types.h file and keep that file identical for diff based validation between the two copies.
Another option would be a new header only library hipMATH for common code.  A dependency that requires no building.  
Troy probably you need to organize a meeting to discuss this, or at the next meeting  @amcamd can add to agenda.

tfalders:
> Troy probably you need to organize a meeting to discuss this, or at the next meeting @amcamd can add to agenda.

Yes, I plan to attend one of the rocBLAS meetings later this week to get the team's input.

daineAMD (approved):
Fine by me as per the meeting. Interesting that there's both [cusolverStatus_t](https://docs.nvidia.com/cuda/cusolver/#id1) and [cublasStatus_t](https://docs.nvidia.com/cuda/cublas/index.html#cublasstatus-t) which are almost identical, but not reused; I suppose they expect(ed) some deviation at some point.

amcamd (approved):

library/include/hipblas.h
```diff
@@ -439,6 +442,28 @@ typedef enum
...
+static_assert(HIPBLAS_FILL_MODE_FULL == 123);
+
+static_assert(HIPBLAS_DIAG_NON_UNIT == 131);
+static_assert(HIPBLAS_DIAG_UNIT == 132);
+
+static_assert(HIPBLAS_SIDE_LEFT == 141);
+static_assert(HIPBLAS_SIDE_RIGHT == 142);
+static_assert(HIPBLAS_SIDE_BOTH == 143);
+
+#endif // __cplusplus
```
cgmb: In C++11, static_assert requires two arguments. The single-argument version was added in C++17.

library/include/hipblas.h
```diff
@@ -398,7 +398,17 @@ typedef enum
...
+typedef enum
+{
+    HIPBLAS_POINTER_MODE_HOST, /**< Scalar values affected by this variable will be located on the host. */
+    HIPBLAS_POINTER_MODE_DEVICE /**<  Scalar values affected by this variable will be located on the device. */
+} hipblasPointerMode_t;
+
 // set the values of enum constants to be the same as those used in cblas
+#ifndef HIPBLAS_SHARED_ENUMS
+#define HIPBLAS_SHARED_ENUMS
+
```
cgmb: I know we discussed this, but I forget the reasoning why we dismissed putting a wrapper around each enum. Then we can just treat all of these enums exactly the same, rather than having some that are different from others. e.g., each one would be like:

```cpp
#ifndef HIPBLAS_FILL_MODE_DECLARED
#define HIPBLAS_FILL_MODE_DECLARED
typedef enum
{
    HIPBLAS_FILL_MODE_UPPER = 121, /**<  Upper triangle */
    HIPBLAS_FILL_MODE_LOWER = 122, /**<  Lower triangle */
    HIPBLAS_FILL_MODE_FULL  = 123
} hipblasFillMode_t;
#elif __cplusplus >= 201103L
static_assert(HIPBLAS_FILL_MODE_UPPER == 121, "inconsistent enum declaration")
static_assert(HIPBLAS_FILL_MODE_LOWER == 122, "inconsistent enum declaration")
static_assert(HIPBLAS_FILL_MODE_FULL == 123, "inconsistent enum declaration")
#endif
```

I recall there was at least one example in which that wasn't useful, but did we decide it was useless in general? We don't need that kind of granularity for SOLVER, but I wonder if it might nevertheless be simpler if all enums are declared in the exact same way.

cgmb (changes requested):
The static assert needs to change, but I also wanted to bring up a question we discussed that I have forgotten the reasoning behind.

library/include/hipblas.h
tfalders: To be honest, I don't recall either. Thinking on it now, however, I think it would be good for future proofing. If hipBLAS and hipSOLVER introduced a new enum in the future with its own wrapper, we can ensure that the enum won't go undefined because, e.g., an older version of a library defines `HIPBLAS_SHARED_ENUMS` without it. I'll make the change.

cgmb (approved):

tfalders:
@amcamd @daineAMD Feel free to merge whenever you wish.

daineAMD:
Lauren is aware of and working on multi-compiler CI issue, I think this PR is fine to be merged without it. @amcamd do you want me to merge?

PR #637, opened 2023-07-18
Doxygen configuration updated to report warnings as errors

rkamd:
resolves # SWDEV-410953

Summary of proposed changes:
- considers doxygen warnings as errors
- Fix doxygen warnings and errors


TorreZuk (approved):
LGTM, contents links on right appear to show correct information now, which is nice.  Same bad link behaviour as rocblas for this panel but unrelated to these improvement.

PR #639, opened 2023-07-20
Add HIPBLAS_USE_HIP_BFLOAT16 compile option to use hip_bfloat16

daineAMD:
This PR adds a compile option `HIPBLAS_USE_HIP_BFLOAT16` to use the `hip_bfloat16` datatype instead of our `hipblasBfloat16` datatype.

Initially I had planned on deprecating the hipblasBfloat16 type and replacing it with the hip_bfloat16 type, but I don't think it's worth it, mainly because hip_bfloat16 doesn't support nvidia backend. hip_bfloat16 is also a pain in that it only gives a minimal definition if not compiling with hipcc.

Users can still get full hipblasBfloat16 class with `HIPBLAS_BFLOAT16_CLASS`, or get minimal definition with no defines.

This adds a simple example showing use of hip_bfloat16, but doesn't add any tests.


TorreZuk (approved):
Seems fine to me, but understanding these type options use cases isn't that clear.
I would put the documentation in the same PR though....

daineAMD:
> Seems fine to me, but understanding these type options use cases isn't that clear. I would put the documentation in the same PR though....

Yeah not sure if anyone would actually be using this, but if hip_bfloat16 becomes mature enough I think it's good to have the option ready/usable.
I'll add some documentation to this PR.

daineAMD:
Added some documentation, happy to have comments to improve it though. I'll make another PR after this one adding documentation for ROCM_MATHLIBS_API_USE_HIP_COMPLEX and HIPBLAS_V2 since we're missing that as well.

docs/functions.rst
```diff
@@ -44,6 +44,15 @@ hipBLAS function uses the following notations to denote precisions:
 - c  = single complex
 - z  = double complex
 
+bfloat 16 Datatype
+==================
+hipBLAS defines a hipblasBfloat16 datatype. This type is exposed as a struct simply containing 16 bits of data. There is also a C++ hipblasBfloat16 class defined
+which gives slightly more functionality, including conversion to and from a 32-bit float datatype. This class can be used by defining HIPBLAS_BFLOAT16_CLASS before including
+the header file hipblas.h.
```
TorreZuk: Since we want to clarify version constraints should add C++ version required for exposure of the class along with the define, e.g. C++11 or greater and defining HIPBLAS_BFLOAT16_CLASS , or similar

clients/samples/example_bfdot_hip_bfloat16.cpp
```diff
@@ -0,0 +1,133 @@
...
+    float gold_resf = bfloat16_to_float(gold_result);
+    float hip_resf  = bfloat16_to_float(hipblas_result);
+    float diff      = std::abs(gold_resf - hip_resf);
+
+    printf("%d    %8.2f        \n", (int)N, gpu_time_used);
+
+    if(diff)
+    {
+        printf("CPU RESULT: %f, GPU_RESULT: %f\n", gold_resf, hip_resf);
+        printf("SSCAL TEST FAILS\n");
```
NaveenElumalaiAMD: sdot?

clients/samples/example_bfdot_hip_bfloat16.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+int main()
+{
+    int             N    = 1024;
+    int             incx = 1;
+    int             incy = 1;
+    hipblasStatus_t status;
+
+    // Naming: dX is in GPU (device) memory. hK is in CPU (host) memory, plz follow this practice
+    std::vector<hip_bfloat16> hx(N);
```
amcamd, line 64: Can you consider adding here
#ifdef HIPBLAS_USE_HIP_BFLOAT16
    //    the hip_bfloat16 usage
#else
    // the hipblasBfloat16 usage
#endif
This would be similar to https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/release/rocm-rel-5.6/clients/samples/example_strmm_v3.cpp#L174 . It would enable the example to be built with or without HIPBLAS_USE_HIP_BFLOAT16 defined. It would also indicate what needs to change when using HIPBLAS_USE_HIP_BFLOAT16 .

amcamd (approved):

PR #640, opened 2023-07-20
Adding compiler options to rmake.py

daineAMD:
- fixes --hip-clang and --no-hip-clang ./install.sh options to use hipcc/g++ accordingly
- adds --hip-clang and --no-hip-clang rmake.py options to use hipcc/g++ accordingly
- uses ROCM_PATH to get hipcc location instead of assuming it's in the path

install.sh defaults to using g++ and rmake.py defaults to using hipcc. I didn't change that in this PR.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -52,7 +52,7 @@ cat <<EOF
 
     --cmake-arg                   Forward the given argument to CMake when configuring the build.
 
-    --compiler </compier/path>    Specify path to host compiler. (e.g. /opt/bin/hipcc)
+    --compiler </compier/path>    Specify path to host compiler. (e.g. /opt/rocm/bin/hipcc)
 
     --custom-target <target>      Specify custom target to link the library against (eg. host, device).
 
@@ -236,7 +236,7 @@ install_packages( )
   if [[ "${build_cuda}" == true ]]; then
     # Ideally, this could be cuda-cublas-dev, but the package name has a version number in it
     library_dependencies_ubuntu+=( "" ) # removed, use --installcuda option to install cuda
-  elif [[ "${build_hip_clang}" == false ]]; then
+  else
     # Custom rocm-dev installation
     if [[ -z ${custom_rocm_dev+foo} ]]; then
       # Install base rocm-dev package unless -v/--rocm-dev flag is passed
@@ -427,7 +427,6 @@ install_prefix=hipblas-install
 build_clients=false
 build_solver=true
 build_cuda=false
-build_hip_clang=true
 build_release=true
 build_relocatable=false
 build_address_sanitizer=false
@@ -497,10 +496,10 @@ while true; do
         build_codecoverage=true
         shift ;;
     --hip-clang)
-        build_hip_clang=true
+        compiler=hipcc
         shift ;;
     --no-hip-clang)
-        build_hip_clang=false
+        compiler=g++
         shift ;;
     --compiler)
         compiler=${2}
--- a/install.sh
+++ b/install.sh
@@ -653,8 +653,6 @@ pushd .
       exit 1
     fi
     cmake_common_options+=("-DBUILD_SHARED_LIBS=OFF")
-    compiler="${rocm_path}/bin/hipcc" #force hipcc for static libs, g++ doesn't work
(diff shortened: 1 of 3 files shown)


install.sh
```diff
@@ -52,7 +52,7 @@ cat <<EOF
 
     --cmake-arg                   Forward the given argument to CMake when configuring the build.
 
-    --compiler </compier/path>    Specify path to host compiler. (e.g. /opt/bin/hipcc)
+    --compiler </compier/path>    Specify path to host compiler. (e.g. /opt/rocm/bin/hipcc)
```
TorreZuk: /compier -> /compiler

rmake.py
```diff
@@ -210,8 +216,20 @@ def config_cmd():
...
+    else:
+        cmake_options.append(f"-DCMAKE_C_COMPILER=gcc")
+        cmake_options.append(f"-DCMAKE_CXX_COMPILER=g++")
+
     if args.static_lib:
         cmake_options.append( f"-DBUILD_SHARED_LIBS=OFF" )
+        # force hipcc for static libs
+        cmake_options.append(f"-DCMAKE_C_COMPILER={rocm_path}/bin/hipcc")
+        cmake_options.append(f"-DCMAKE_CXX_COMPILER={rocm_path}/bin/hipcc")
+        print("Forcing compiler to hipcc for static library.")
```
TorreZuk: Can you document the reason for this?

rmake.py
```diff
@@ -210,8 +216,20 @@ def config_cmd():
     create_dir( os.path.join(build_path, "clients") )
     os.chdir( build_path )
 
+    # compiler
+    if args.use_hipcc_compiler:
+        cmake_options.append(f"-DCMAKE_C_COMPILER={rocm_path}/bin/hipcc")
+        cmake_options.append(f"-DCMAKE_CXX_COMPILER={rocm_path}/bin/hipcc")
+    else:
+        cmake_options.append(f"-DCMAKE_C_COMPILER=gcc")
+        cmake_options.append(f"-DCMAKE_CXX_COMPILER=g++")
```
TorreZuk: Sorry I can't recall how this builds on windows but would some or all of this be in a not os "nt" case?

TorreZuk (approved):
Fine but expect fixes for windows are required.

rmake.py
daineAMD: Seems to build fine with g++ for me now, so not sure why I put this originally. I'll do some searching through emails/jira and see if I come up with anything.

rmake.py
daineAMD: I see it's set to clang++ in [toolchain-windows.cmake](https://github.com/ROCmSoftwarePlatform/hipBLAS/blob/develop/toolchain-windows.cmake#L13), so it takes clang++ regardless, is that right?
Either way, I'll fiddle around on windows and update accordingly.

TorreZuk: Well if you know it get's overridden it may confuse to explicitly set on command line, did windows CI run ?

daineAMD: CI ran and passed. I updated it a bit to note that hipblas currently can't be build without hip-clang. Can try building hipBLAS with other compilers on windows at some point if we decide it's worth pursuing.

rmake.py
daineAMD: I removed this since it seems to work.

rmake.py
```diff
@@ -210,6 +216,17 @@ def config_cmd():
     create_dir( os.path.join(build_path, "clients") )
     os.chdir( build_path )
 
+    # compiler
+    if args.use_hipcc_compiler:
+        cmake_options.append(f"-DCMAKE_C_COMPILER={rocm_path}/bin/hipcc")
+        cmake_options.append(f"-DCMAKE_CXX_COMPILER={rocm_path}/bin/hipcc")
+    else:
+        if os.name == "nt":
+            print("Currently forcing hip-clang compiler on Windows.")
```
TorreZuk: I meant guard out both cases as even the top ones aren't tested on windows (could be a linux only option?), hipcc.bat is not the same as the clang.exe we use in toolchain so seeing that is misleading.  If the build shows it actually is used then that is also fine.  There was even a hipcc.exe at one point but that is in the same case.

PR #641, opened 2023-07-24
Adding default to some switches to avoid compiler warnings

daineAMD:


TorreZuk (approved):
You could comment to briefly say you are breaking the pattern of letting compiler warn as there are many hip types that are invalid that vary by windows/linux ?

daineAMD:
Builds fine locally. Low risk and a small change so will merge even though CI is still broken.

PR #644, opened 2023-07-26
use test type tolerances from rocBLAS

TorreZuk:
* client side only changes to fix 16bit test failures seen on gfx1100
* taken from rocBLAS which are used with larger test sizes so could be tightened later


daineAMD (approved):

TorreZuk:
Verbal approval from Andrew so merging

PR #632, opened 2023-07-13
add new APIs *WithFlags

jeffdaily:
- hipblasGemmExWithFlags
- hipblasGemmBatchedExWithFlags
- hipblasGemmStridedBatchedExWithFlags

Implements SWDEV-410872.

PR #645, opened 2023-07-27
Adding fortran samples, fixing some of the fortran interface.

daineAMD:
Fixes parts of the fortran interface:
- removes unused params from hipblasGemmEx interfaces (as prompted by #632, I can deal with merge conflicts). I think that the compiler is fine with the extra parameters when interoperating with C++ interface from clients and ends up just ignoring the extra paremeters, but not sure.
- Fixes HIPBLAS_GEMM_DEFAULT value in fortran module. I think it's set correctly in C++ test suite so is passed on ok. Caught the error with fortran sample code
- Adds (already created) hipblasSscal fortran example
- Adds hipblasGemmEx fortran example

I'm not sure if these sorts of issues can be caught with the test suite as it is, or if we need some test code in fortran. Whether or not that is worth it while the fortran is handed off is another issue.


TorreZuk (approved):
Looks good to me.  Good to have a sample now too

PR #647, opened 2023-07-28
Swdev 410872 gemm Ex with flags

TorreZuk:
* taking over from Jeff's work in PR #632
* will rebase off Daine's PR once he merges and continue 
* @nielenventer you can please review last commit only (solution_index drop)


nielenventer (approved):

library/include/hipblas.h
```diff
@@ -18159,6 +18283,32 @@ HIPBLAS_EXPORT hipblasStatus_t hipblasGemmStridedBatchedEx_v2(hipblasHandle_t
...
+                                            const void*          beta,
+                                            void*                C,
+                                            hipDataType          cType,
+                                            int                  ldc,
+                                            hipblasStride        strideC,
+                                            int                  batchCount,
+                                            hipblasComputeType_t computeType,
+                                            hipblasGemmAlgo_t    algo,
+                                            hipblasGemmFlags_t   flags);
+
```
daineAMD, line 18311: Needs `#ifdef HIPBLAS_V2 #define hipblasGemmWithFlags hipblasGemmWithFlags_v2` around L20624 to keep with the pattern.

library/src/hipblas_module.f90
```diff
@@ -12053,17 +12053,48 @@ function hipblasGemmEx(handle, transA, transB, m, n, k, alpha, a, a_type, lda, &
             integer(c_int), value :: ldc
             integer(kind(HIPBLAS_R_16F)), value :: compute_type
             integer(kind(HIPBLAS_GEMM_DEFAULT)), value :: algo
+        end function hipblasGemmEx
+    end interface
+
+    interface
+        function hipblasGemmExWithFlags(handle, transA, transB, m, n, k, alpha, a, a_type, lda, &
+                                        b, b_type, ldb, beta, c, c_type, ldc, &
+                                        compute_type, algo, solution_index, flags) &
```
daineAMD: WithFlags version doesn't need solution_index here either.

library/src/hipblas_module.f90
```diff
@@ -12089,17 +12120,49 @@ function hipblasGemmBatchedEx(handle, transA, transB, m, n, k, alpha, a, a_type,
...
+            integer(kind(HIPBLAS_R_16F)), value :: b_type
+            integer(c_int), value :: ldb
+            type(c_ptr), value :: beta
+            type(c_ptr), value :: c
+            integer(kind(HIPBLAS_R_16F)), value :: c_type
+            integer(c_int), value :: ldc
+            integer(c_int), value :: batch_count
+            integer(kind(HIPBLAS_R_16F)), value :: compute_type
+            integer(kind(HIPBLAS_GEMM_DEFAULT)), value :: algo
             integer(c_int32_t), value :: solution_index
```
daineAMD: Same here

TorreZuk: Sorry @daineAMD didn't really start on this yet, only wanted Nielen to sign off on deletion of solution index.  I'll push a partial implementation of testing to get your review before going on to batched & strided with it.  Only v2 could have the new API from my understanding so there is no inconsistency...

TorreZuk:
Okay @daineAMD now ready for your initial review it is still WIP but has proposed form of testing can be reviewed.  This needs a rebase after your PR and so if you have time you can also take over and continue next week if you have cycles to spare.

daineAMD:
@jeffdaily This PR adds gemmEx*WithFlags functions as you proposed in #632. Some changes compared to your PR:
- adds testing of new functions
- doesn't include version with hipblasDatatype_t or follow the same *_v2 pattern.

As we discussed, it seems like you're OK with having this break the *_v2 pattern. Just to be clear, this means that the added functions use `hipDataType` instead of `hipblasDatatype_t` as the rest of the hipBLAS API uses (deprecated in the next release). It also uses `hipblasComputeType_t` for the `compute_type` instead of `hipblasDatatype_t`. Thus, if you're not defining HIPBLAS_V2, you will have to deal with these differences until the old interfaces are removed and all uses of hipblasDatatype_t will have to be updated. I wanted to double check that this is OK with your team.

Please take a look at the updated API in hipblas.h when you have a chance, you can ignore any changes on the clients/ side.

jeffdaily:
How hard is it to have both the V1 and V2 *WithFlags implementations?  pytorch and onnxruntime are both still using V1 hipblas APIs, and I haven't tried using V2 APIs yet.  Further, pytorch is also integrating with hipsolver and hipblaslt libraries -- will these other libraries have V2 APIs?  It's unknown to me whether switching to V2 APIs will be difficult.  Which hipblas version introduces the V2 APIs and the HIPBLAS_V2 macro, and which version of hipblas will remove the V1 APIs?

daineAMD:
> How hard is it to have both the V1 and V2 *WithFlags implementations? pytorch and onnxruntime are both still using V1 hipblas APIs, and I haven't tried using V2 APIs yet. Further, pytorch is also integrating with hipsolver and hipblaslt libraries -- will these other libraries have V2 APIs? It's unknown to me whether switching to V2 APIs will be difficult. Which hipblas version introduces the V2 APIs and the HIPBLAS_V2 macro, and which version of hipblas will remove the V1 APIs?

It isn't difficult to add both implementations; if the V1 versions will be used then I don't mind adding them. The motivation to not add the hipblasDatatype_t APIs is that they will me immediately deprecated when added. But if the overhead of incorporating the hipDataType/hipblasComputeType_t for these specific APIs is too much, it's probably easier to just add both versions in hipBLAS. In general I don't think it will be too difficult to change over to V2 APIs as hipblasDatatype_t maps fairly easily to hipDataType, though.

hipSOLVER isn't using any types in hipBLAS which are being deprecated as far as I can tell, so the V2 changes in hipBLAS shouldn't correspond to any changes there. I don't have any experience with hipblaslt, but I see that they are using our recently deprecated types so I will work with them on changing over, how this ends up being done and how it impacts their API isn't clear right now.

HIPBLAS_V2 macro is **not** in ROCm 5.7, it's just been introduced recently so will make it into the next ROCm release after 5.7 (same as the changes in this PR). The tentative plan for removing V1 APIs would be the next major ROCm release, i.e. ROCm 7.0.

jeffdaily:
Then I request that the V1 API gemm ex flag be added, as well.  That will give us more time to work on the transition to the V2 APIs.  Thanks.

library/include/hipblas.h
```diff
@@ -17765,6 +17826,48 @@ HIPBLAS_EXPORT hipblasStatus_t hipblasGemmEx_v2(hipblasHandle_t      handle,
                                                 hipblasComputeType_t computeType,
                                                 hipblasGemmAlgo_t    algo);
 
+HIPBLAS_EXPORT hipblasStatus_t hipblasGemmExWithFlags(hipblasHandle_t    handle,
```
amcamd: I think you need a HIPBLAS_DEPRECATED_MSG here, and for the _batched and _strided_batched. This would be similar to the ROCBLAS_DEPRECATED_MSG here
https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/release/rocm-rel-5.7/library/include/internal/rocblas-functions.h#L13004

amcamd (commented):
I think you need to add a deprecation notice in the file: https://github.com/ROCmSoftwarePlatform/hipBLAS/blob/develop/docs/deprecation.rst

daineAMD:
> I think you need to add a deprecation notice in the file: https://github.com/ROCmSoftwarePlatform/hipBLAS/blob/develop/docs/deprecation.rst

Sure, I can add this function to the list as well once #649 goes in.

library/include/hipblas.h
daineAMD: I only added HIPBLAS_DEPRECATED_MSG for the top-level hipblasDatatype_t, instead of adding it for each function which uses hipblasDatatype_t to avoid excessive warnings. We can discuss if there's a reason to add it for each function as well.

amcamd: OK, deprecation message for top-level hipblasDatatype_t is good.

amcamd (approved):

clients/gtest/gemm_ex_gtest.cpp
```diff
@@ -709,7 +743,8 @@ INSTANTIATE_TEST_SUITE_P(quick_blas_batched_ex_small_hpa_half_single_out,
                                  ValuesIn(precision_hpa_half_single_out),
                                  ValuesIn(compute_hpa_half_single_out),
                                  ValuesIn(batch_count_range),
-                                 ValuesIn(is_fortran)));
+                                 ValuesIn(is_fortran),
+                                 ValuesIn(without_flags)));
```
TorreZuk: Catching up, could this one be with_and_without ?  And is strided now tested as looks like you completed the full implementation now, right @daineAMD ?  I'll leave this to you as you did more than me now

daineAMD: I'm guessing that Tensile would support this precision config as well but not sure, I added it in another commit so we can see how it behaves.
Yes, strided_batched (along with batched) is tested, full implementation is complete so should be good to go, will wait for another CI run.

daineAMD:
@jeffdaily Do the changes in hipblas.h look good to you? If so I'll go ahead and merge this.

TorreZuk (commented):
LGTM it won't let me approve.

jeffdaily (approved):

PR #651, opened 2023-08-02
Fix install when not building hipblas-bench

cgmb:
The hipblas-common target is only used by hipblas-bench and will therefore not be built during a build of only the library and tests. In that case, the install will fail.

This fixes the error by not installing hipblas-common unless hipblas-bench is to be built.

### DIFF
--- a/clients/CMakeLists.txt
+++ b/clients/CMakeLists.txt
@@ -135,13 +135,15 @@ add_custom_command( OUTPUT "${HIPBLAS_GENTEST}"
 
 add_custom_target( hipblas-common DEPENDS "${HIPBLAS_COMMON}" "${HIPBLAS_TEMPLATE}" "${HIPBLAS_SMOKE}" "${HIPBLAS_GENTEST}" )
 
-rocm_install(
-  FILES ${HIPBLAS_COMMON} ${HIPBLAS_TEMPLATE} ${HIPBLAS_SMOKE}
-  DESTINATION "${CMAKE_INSTALL_BINDIR}"
-  COMPONENT clients-common
-)
-rocm_install(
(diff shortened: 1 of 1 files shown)


daineAMD (approved):
This won't be tested by CI, so approving assuming you've done the appropriate testing locally, I can try it out as well but LGTM.

clients/CMakeLists.txt
```diff
@@ -135,13 +135,15 @@ add_custom_command( OUTPUT "${HIPBLAS_GENTEST}"
...
-  DESTINATION "${CMAKE_INSTALL_BINDIR}"
-  COMPONENT clients-common
-)
-rocm_install(
-  PROGRAMS ${HIPBLAS_GENTEST}
-  DESTINATION "${CMAKE_INSTALL_BINDIR}"
-  COMPONENT clients-common
-)
+if(BUILD_CLIENTS_BENCHMARKS)
+  rocm_install(
```
TorreZuk, line 139: This looks wrong, if guarding it should be if( BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS) etc.

TorreZuk (changes requested):
Guard needs to be fixed

clients/CMakeLists.txt
cgmb: hipblas-test is not marked as depending on clients-common, so that would cause the build to fail.

TorreZuk: Probably can fix it all up after samples is merged

TorreZuk:
Added deps on common and this change in linked PR so closing this one.

PR #656, opened 2023-08-10
Deprecating hipblasComplex types and adding _v2 version of ixamax

daineAMD:
Beginning of deprecation of hipblas complex types. Just made changes for ixamax for now, please let me know if you had something else in mind for this process.


docs/functions.rst
```diff
@@ -79,6 +79,9 @@ std::complex<float>, and hipblasComplex, as well as for the double precision var
 users can use either a compiler define or inline #define ROCM_MATHLIBS_API_USE_HIP_COMPLEX before including the header file <hipblas.h>. Thus, the
 API is compatible with both forms, but recompilation is required to avoid casting if switching to pass in the hip complex types.
 
+Note that hipblasComplex, hipblasDoubleComplex, and use of ROCM_MATHLIBS_API_USE_HIP_COMPLEX are now deprecated. The API will provide interfaces
+using hipComplex and hipDoubleComplex only in the future.
```
TorreZuk: Suggestion: Using "only" before the types may be clearer.

clients/gtest/blas1_gtest.cpp
```diff
@@ -2286,6 +2300,24 @@ TEST_P(blas1_gtest, amax_batched_float_complex)
...
+
+    if(status != HIPBLAS_STATUS_SUCCESS)
+    {
+        if(arg.batch_count < 0)
+        {
+            EXPECT_EQ(HIPBLAS_STATUS_INVALID_VALUE, status);
+        }
+        else
+        {
+            EXPECT_EQ(HIPBLAS_STATUS_SUCCESS, status); // fail
```
TorreZuk, line 2316: All the // fail comments not really sure of.

TorreZuk (approved):
LGTM seems to match the other v2 pattern.  Minor suggestion made

clients/gtest/blas1_gtest.cpp
daineAMD: Yeah, that line just ensures that we actually get a gtest failure when we don't get HIPBLAS_STATUS_SUCCESS from testing_*.hpp. `// fail` is meant to imply that if we get to that line, the test will fail. hipBLAS gtest framework needs some rework.

PR #659, opened 2023-08-15
Fixing fortran sample code

daineAMD:


TorreZuk (commented):
So the fixes are for flang or other fortran?  But you are setting to gfortran now in install.

daineAMD:
> So the fixes are for flang or other fortran? But you are setting to gfortran now in install.

Fixes are for flang - should build fine with flang or gfortran now. I just moved the FC=gfortran from the deps installation since you'd get different behaviour if installing deps vs. not installing deps with a different compiler. Not sure if we want to add "support" for flang or not, but maybe would be best to remove setting FC in the install script (it isn't set it rmake) and let CMake default to gfortran if not set.

amcamd (approved):
LGTM

PR #661, opened 2023-08-17
hipComplex Versions of L1 Functions

daineAMD:
Adding hipComplex/hipDoubleComplex version for all L1 functions.

Details:
- hipblas_template_specialization.cpp changes: We're using hipblasComplex types for clients, so need to cast to hipComplex well calling hipblas.h API when using HIPBLAS_V2 define.
- blas1_gtest.cpp changes: adding tests for single/double complex precisions
- hipblas_fortran.hpp changes: client-side C++ interface for Fortran interface, changes needed to use new API for HIPBLAS_V2 changes for minimal code changes in test infrastructure. No changes made in actual .f90/.mod files
- hipblas.h changes: introduce new _v2 API, add `#define` when HIPBLAS_V2 is defined

### DIFF
--- a/clients/common/hipblas_template_specialization.cpp
+++ b/clients/common/hipblas_template_specialization.cpp
@@ -84,7 +84,12 @@ hipblasStatus_t hipblasAxpy<hipblasComplex>(hipblasHandle_t       handle,
                                             hipblasComplex*       y,
                                             int                   incy)
 {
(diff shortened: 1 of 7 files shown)


amcamd (approved):
LGTM
- checked that there is a deprecation notice in hipBLAS/docs/deprecation.rst
- this is for SWDEV-414800

PR #662, opened 2023-08-21
hipComplex Versions of L2 Functions

daineAMD:
Same as #661 but for L2 functions. Lots of line changes so thank you in advance for taking the time to review!


amcamd (approved):
LGTM

PR #670, opened 2023-09-06
Fix --yaml flag for hipblas-bench and hipblas-test

daineAMD:
--yaml flag for hipblas-bench/hipblas-test is likely broken since #647.

With this change `./hipblas-test --yaml hipblas_smoke.yaml` all passed, `./hipblas-bench --yaml hipblas_smoke.yaml` works as well.

### DIFF
--- a/.jenkins/common.groovy
+++ b/.jenkins/common.groovy
@@ -47,6 +47,13 @@ def runTestCommand (platform, project)
                 """
 
     platform.runCommand(this, v2TestCommand)
+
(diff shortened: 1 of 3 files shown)


TorreZuk (approved):
LGTM, wonder if you can add a run of the smoke test to the groovy so we get the --yaml tested in CI?

TorreZuk:
Also want to ffwd this to release-staging I expect

eidenyoshida (approved):

PR #676, opened 2023-09-14
Use yaml based testing 1

daineAMD:
This introduces yaml based testing to hipBLAS as we use in rocBLAS. This will make it easier to ensure that we have good coverage (we don't currently) and make it easier to add/remove tests as needed. It will also add the ability to easily add tests for just AMD/Nvidia backend. It includes most but not all features currently in the rocblas client code. This won't be going into ROCm 6.0, I may close and re-open after some review + CI. We can discuss having different branches for ROCm 6.0/6.1 alternatively.

This will be the first in a series of PRs. After this, I will translate all the current test cases to yaml, then remove old test infrastructure code, and finally add further tests to ensure we have full coverage. I'll also add documentation, likely in this PR or a separate PR.


clients/gtest/blas1/asum_gtest.cpp
```diff
@@ -0,0 +1,158 @@
...
+                    name << "_" << arg.batch_count;
+                }
+            }
+
+            if(arg.api == FORTRAN)
+            {
+                name << "_F";
+            }
+
+            return std::move(name);
```
TorreZuk: Always thought if we were starting from scratch I would do this instead with a class much like ArgumentModel logging although maybe not variadic constructor.  As you are starting from scratch, might be worth a look although following rocblas also has it's benefits.

clients/include/testing_asum.hpp
```diff
@@ -37,7 +37,7 @@ inline void testname_asum(const Arguments& arg, std::string& name)
 }
 
 template <typename T>
-inline hipblasStatus_t testing_asum(const Arguments& arg)
+inline void testing_asum(const Arguments& arg)
```
TorreZuk: Was never sure why these were inline (pre-existing), compiler may be ignoring anyway

clients/include/testing_asum.hpp
```diff
@@ -54,18 +54,19 @@ inline hipblasStatus_t testing_asum(const Arguments& arg)
         device_vector<Tr> d_hipblas_result_0(1);
         host_vector<Tr>   h_hipblas_result_0(1);
         hipblas_init_nan(h_hipblas_result_0.data(), 1);
-        CHECK_HIP_ERROR(
+        CHECK_HIP_ERROR_NO_RETURN(
             hipMemcpy(d_hipblas_result_0, h_hipblas_result_0, sizeof(Tr), hipMemcpyHostToDevice));
 
-        CHECK_HIPBLAS_ERROR(hipblasSetPointerMode(handle, HIPBLAS_POINTER_MODE_DEVICE));
-        CHECK_HIPBLAS_ERROR(hipblasAsumFn(handle, N, nullptr, incx, d_hipblas_result_0));
+        CHECK_HIPBLAS_ERROR_NO_RETURN(hipblasSetPointerMode(handle, HIPBLAS_POINTER_MODE_DEVICE));
```
TorreZuk: Maybe shorter macro name ?    ASSERT_HIP_SUCCESS( )  maybe no better.

TorreZuk (approved):
LGTM, will be much simpler to extend test coverage.   Is there a ticket?

daineAMD:
Thought I had already made a ticket but I guess not, created LWPMLSE-486.

TorreZuk:
@daineAMD won't this go into develop-6.1 ?  If so you can merge to develop-6.1 and I can help cover some functions if you want.

TorreZuk:
Not sure but @eidenyoshida does the CI support dependencies when building on develop-6.1  for hipBLAS?

daineAMD:
If CI support isn't 100% for develop-6.1 there no problem with me leaving it against develop until right before merge.

clients/gtest/blas1/asum_gtest.cpp
daineAMD: I think this is a good idea, I'll draft something up in another PR after this goes in.

TorreZuk: > I think this is a good idea, I'll draft something up in another PR after this goes in.

I think I already wrote one for hipBLAS I had just forgot  :-) ,  rocBLAS we don't have one yet
class ArgumentModel
    void test_name(const Arguments& arg, std::string& name)
    
I was using that for the yaml tests I added for smoke testing I presume.... maybe you just need to hook into that form.

daineAMD: Ha, even better; I'll make some further changes here then.

clients/gtest/blas1/asum_gtest.cpp
```diff
@@ -62,49 +62,22 @@ namespace
...
-
-            if(arg.api == FORTRAN)
-            {
-                name << "_F";
-            }
-
-            return std::move(name);
+            std::string name;
+            testname_asum(arg, name);
+            return name;
```
TorreZuk: return std::move(name);
?  Did you find it didn't avoid a copy

daineAMD: No, std::move should still be useful, just overlooked. Added back.

daineAMD:
CI is having some problems, but looks like all pipelines have passed at some point:
[pre-checkin](http://math-ci.amd.com/blue/organizations/jenkins/main%2Fprecheckin%2FhipBLAS/detail/PR-676/1/pipeline)
[cuda](http://math-ci.amd.com/blue/organizations/jenkins/main%2Fprecheckin-cuda%2FhipBLAS/detail/PR-676/5/pipeline)
[windows](http://mlsejenkinsvm.amd.com/blue/organizations/jenkins/ROCmMathLib%2FhipBLAS/detail/PR-676/5/pipeline)
[static-analysis](http://math-ci.amd.com/blue/organizations/jenkins/main%2Fstatic-analysis%2FhipBLAS/detail/PR-676/5/pipeline)

Recent pre-checkin failures due to rocsparse dependency CI problems, recent cuda failures due to updated cuda image (some discussion in Teams channel).

Going to merge this so easier to move forward with other PRs relating to this but will keep an eye out on CI once [rocSPARSE passes CI](http://10.216.51.142/blue/organizations/jenkins/main%2Fprecheckin%2FrocSPARSE/detail/develop/612/pipeline) and cuda pipeline is fixed.

PR #677, opened 2023-09-15
Prepend rocblas/rocsolver custom include directories

daineAMD:
When using the `--rocblas-path` or `--rocsolver-path` options while having an installed rocblas/rocsolver on your system, the build system would have the `/opt/rocm/include` directory in the cmd line before `/my/custom/rocblas/build/rocblas-install`, thus it would include the incorrect rocblas/rocsolver.h file. This should fix that by prepending the include directory list if using these commands.

### DIFF
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -86,6 +86,12 @@ if( NOT USE_CUDA )
       if( CUSTOM_ROCSOLVER)
         set ( ENV{rocsolver_DIR} ${CUSTOM_ROCSOLVER})
         find_package( rocsolver REQUIRED CONFIG NO_CMAKE_PATH )
+
+        # in case of using custom rocsolver and not custom rocblas, we need to have
+        # custom rocsolver include directories before rocblas/hip include directories
+        # in case there is a rocsolver installed on the system.
+        target_include_directories( hipblas
+          SYSTEM PRIVATE $<BUILD_INTERFACE:${ROCSOLVER_INCLUDE_DIRS}> )
       elseif(WIN32)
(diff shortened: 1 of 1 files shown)


NaveenElumalaiAMD (approved):
Thanks for the fix!

library/src/CMakeLists.txt
```diff
@@ -76,6 +76,8 @@ if( NOT USE_CUDA )
     if( CUSTOM_ROCBLAS )
       set ( ENV{rocblas_DIR} ${CUSTOM_ROCBLAS})
       find_package( rocblas REQUIRED CONFIG NO_CMAKE_PATH )
+      target_include_directories( hipblas BEFORE PRIVATE ${CUSTOM_ROCBLAS}/include)
+
```
TorreZuk: Since you did the find_package what about just using ROCBLAS_INCLUDE_DIR and putting it in the list further down below.  You are making assumptions about the include that should be covered in the rocblas-config

library/src/CMakeLists.txt
```diff
@@ -76,6 +76,8 @@ if( NOT USE_CUDA )
     if( CUSTOM_ROCBLAS )
       set ( ENV{rocblas_DIR} ${CUSTOM_ROCBLAS})
       find_package( rocblas REQUIRED CONFIG NO_CMAKE_PATH )
+      target_include_directories( hipblas BEFORE PRIVATE ${CUSTOM_ROCBLAS}/include)
```
TorreZuk: set( ENV ... why wasn't the path just passed into the find_package call ?
Unrelated but just curious.

TorreZuk (approved):
LG but suggest an alternative that may or may not be more robust.

TorreZuk:
Yes that common /rocm/include leads to some ugly support code

PR #678, opened 2023-09-19
replace build.image with build.os in the .readthedocs.yaml config

samjwu:
resolves: update ReadtheDocs configuration
build.image (currently used implicitly) will be deprecated in October 2023
https://docs.readthedocs.io/en/stable/config-file/v2.html

Summary of proposed changes:
- replace build.image with build.os in the .readthedocs.yaml config

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -9,6 +9,10 @@ sphinx:
 formats: [htmlzip]
 
 python:
-   version: "3.8"
    install:
    - requirements: docs/.sphinx/requirements.txt
+
+build:
+   os: ubuntu-20.04
+   tools:
+      python: "3.8"


.readthedocs.yaml
```diff
@@ -9,6 +9,10 @@ sphinx:
 formats: [htmlzip]
 
 python:
-   version: "3.8"
    install:
    - requirements: docs/.sphinx/requirements.txt
+
+build:
+   os: ubuntu-20.04
```
TorreZuk, line 16: This won't affect me if I build locally with Ubuntu22 will  it?

TorreZuk (approved):
Seems okay

.readthedocs.yaml
samjwu: No, this is just what the read the docs server uses.

PR #681, opened 2023-09-22
Updating CUDA hip package and path

daineAMD:
Into develop for 6.0


TorreZuk (approved):
Worth a try.  So how is it different than !USE_CUDA ?  same find_package could work?

daineAMD:
> Worth a try. So how is it different than !USE_CUDA ? same find_package could work?

Ah yeah if this works then probably can use same path.

daineAMD:
CI passed so merging. I'll leave the cleanup mentioned for another PR so we can get this into FC today.

PR #684, opened 2023-09-27
Update lib target soversion format

arvindcheru:
resolves #___

Summary of proposed changes:
- Update lib target so version format to recommended format - (ex: libhipblas.so.2.0.0. updated to libhipblas.so.2.0.)
-
-


TorreZuk (approved):
LGTM

PR #688, opened 2023-09-28
Simplify find hip package control flow

daineAMD:
Might as well get this in 6.0 FC


TorreZuk (approved):
Yes if that is 6.0 way, now if you want to build code for 6.0 against 5.7 rocm we get into that discussion as this will not work, right?

daineAMD:
New commit should support building hipblas 6.0 with hip/rocm < 6.0 on cuda, let me know what you think.

TorreZuk (approved):
LGTM

PR #689, opened 2023-09-28
Yaml testing for all L2 functions

daineAMD:


clients/gtest/blas2/gbmv_gtest.cpp
```diff
@@ -0,0 +1,147 @@
...
+            {
+            case GBMV:
+                return !strcmp(arg.function, "gbmv") || !strcmp(arg.function, "gbmv_bad_arg");
+            case GBMV_BATCHED:
+                return !strcmp(arg.function, "gbmv_batched")
+                       || !strcmp(arg.function, "gbmv_batched_bad_arg");
+            case GBMV_STRIDED_BATCHED:
+                return !strcmp(arg.function, "gbmv_strided_batched")
+                       || !strcmp(arg.function, "gbmv_strided_batched_bad_arg");
+            }
```
TorreZuk, line 77: Just an idea, maybe a bad one, but not sure if the compiler can optimize out the switch:
but all can just be:
if constexpr ( (a) == (b) )
Can use in name_suffix too, as the else is not required so a define IF_GBMV_EQ( type ), but that is probably not as readable.

TorreZuk (approved):
Very good to see! my suggestion only for future consideration.  I hope you have a script doing a bunch of the repetitive pattern.  Would think some new base class could be injected to force more of the pattern but the templating gets in the way so no good suggestion from me.

PR #690, opened 2023-09-28
Change some L2 apis to use 'n' instead of 'm' where applicable

daineAMD:
Changing some L2 functions (with triangular and symmetric matrices) to use `n` in signature for size instead of `m` to match legacy blas.


TorreZuk (approved):
If you prefer this is fine by me.   Is this for legacy blas consistency?  rocBLAS would get similar renaming where applicable?

daineAMD:
> If you prefer this is fine by me. Is this for legacy blas consistency? rocBLAS would get similar renaming where applicable?

Yeah consistency with both legacy blas and cuBLAS. Would be good to have rocBLAS match too but not important, just something I stumbled across.

daineAMD:
Looks like CI might not be enabled for cuda pipeline for develop-6.1 PRs. Ran manually on cuda machine and built successfully and passed all blas2 tests.

PR #692, opened 2023-10-02
Yaml testing for all L3 functions

daineAMD:


TorreZuk (approved):
Excellent to see.  Didn't spot anything beyond the first two phases.

TorreZuk:
If you get blocked on CI you could do a partial bump for BLAS if waiting on solver.  Same goes for FC the versions are known as no commit right?  Could proceed and do partial on develop but I would just ping solver team.

daineAMD:
hbmv failure on gfx90a but unrelated to this PR, could be related to L2 PR, will see if I can reproduce and if it shows up on CI again.

PR #705, opened 2023-10-16
blas_ex yaml testing

daineAMD:


clients/include/hipblas_common.yaml
```diff
@@ -141,6 +141,14 @@ Real precisions: &real_precisions
...
   - &hpa_bf16_precision
     { a_type:  bf16_r, b_type:  bf16_r, c_type: bf16_r, d_type: bf16_r, compute_type: f32_r }
+  - &alpha_f32_hpa_bf16_precision
+    { a_type:  f32_r, b_type:  bf16_r, c_type: bf16_r, d_type: bf16_r, compute_type: f32_r }
+  - &hpa_half_float_alpha
+    { a_type: f32_r, b_type: f16_r, c_type: f16_r, d_type: f16_r, compute_type: f32_r }
+  - &hpa_bf16_in_single_out_precision
+    { a_type:  bf16_r, b_type:  bf16_r, c_type: f32_r, d_type: f32_r, compute_type: f32_r }
+  - &hpa_half_in_single_out_precision
+    { a_type: f16_r, b_type: f16_r, c_type: f32_r, d_type: f32_r, compute_type: f32_r }
```
TorreZuk, line 151: I think these names are confusing same as rocblas. alpha matches compute only sometimes?  Wonder if should standardize but can be later PR.

clients/include/type_dispatch.hpp
```diff
@@ -135,18 +135,32 @@ auto hipblas_blas1_ex_dispatch(const Arguments& arg)
...
+    {
+        // bfloat16 scal, nrm2
+        return TEST<hipblasBfloat16, hipblasBfloat16, float>{}(arg);
+    }
+    else if(is_axpy && Ta == Tex && Tx == Ty && (Tx == HIPBLAS_R_16B || Tx == HIPBLAS_R_16F)
+            && Tex == HIPBLAS_R_32F)
+    {
+        // axpy bfloat16 with float alpha
+        return TEST<float, hipblasBfloat16, hipblasBfloat16, float>{}(arg);
+    }
```
TorreZuk, line 153: This pattern too probably should be improved but fine for now.

clients/include/utility.h
```diff
@@ -211,6 +211,26 @@ struct real_t_impl<T, std::enable_if_t<is_complex<T>>>
...
+constexpr double hipblas_type_epsilon<hipblasComplex> = std::numeric_limits<float>::epsilon();
+template <>
+constexpr double
+    hipblas_type_epsilon<hipblasDoubleComplex> = std::numeric_limits<double>::epsilon();
+template <>
+constexpr double
+    hipblas_type_epsilon<hipblasHalf> = 1.0009765625; // fp16 0x3C01 in double precision
+template <>
+constexpr double
+    hipblas_type_epsilon<hipblasBfloat16> = 1.0078125; // bf16 0x3F81 in double precision
```
TorreZuk: Epsilon I don't expect 1.0 + eps, doesn't match the pattern.

clients/include/utility.h
```diff
@@ -440,6 +460,13 @@ inline hipblasBfloat16 random_hpl_generator()
         float_to_bfloat16(std::uniform_real_distribution<float>(-0.5, 0.5)(hipblas_rng)));
 }
 
+template <>
+inline hipblasHalf random_hpl_generator()
+{
+    return hipblasHalf(
+        float_to_half(std::uniform_real_distribution<float>(-0.5, 0.5)(hipblas_rng)));
```
TorreZuk: Didn't look at actual instantiation options but -0.5f, 0.5f would be clearer.

TorreZuk (approved):
Only epsilon question concerns me, other mentions are future work or can be ignored.

clients/include/hipblas_common.yaml
daineAMD: Yeah it's a mess in there, would be good to organize and/or standardize. Yes rocblas backend supports alpha_type != compute_type for some functions, don't think cuBLAS ever supports that.

clients/include/utility.h
daineAMD: Good catch thanks.

clients/include/utility.h
```diff
@@ -222,14 +222,14 @@ constexpr double hipblas_type_epsilon<double> = std::numeric_limits<double>::eps
...
 template <>
-constexpr double hipblas_type_epsilon<hipblasDoubleComplex>
-    = std::numeric_limits<double>::epsilon();
+constexpr double
+    hipblas_type_epsilon<hipblasDoubleComplex> = std::numeric_limits<double>::epsilon();
 template <>
-constexpr double hipblas_type_epsilon<hipblasHalf>
-    = 0.0009765625; // diff between 1 and fp16 0x3C01 in double precision
+constexpr double hipblas_type_epsilon<
+    hipblasHalf> = 0.0009765625; // diff between 1 and fp16 0x3C01 in double precision
```
TorreZuk: Sorry to be picky but the comment is a bit cryptic, maybe "in fp16 diff between 0x3C00 (1.0) and 0x3C01 ...."

PR #707, opened 2023-10-19
Solver and aux yaml based testing

daineAMD:
This should be it. Will have 1 more PR to clean up loose ends.


clients/include/aux/testing_set_get_matrix.hpp
```diff
@@ -54,7 +54,7 @@ inline hipblasStatus_t testing_set_get_matrix(const Arguments& arg)
     // memory
     if(rows < 0 || cols < 0 || lda <= 0 || ldb <= 0 || ldc <= 0)
     {
-        return HIPBLAS_STATUS_INVALID_VALUE;
+        return;
     }
```
TorreZuk, line 58: Maybe need a ticket or TODO as in the future this should be a bad_arg test or report arguments for which no test was run so the yaml can be cleaned.

TorreZuk (approved):
Excellent, solver may even want to use this to run some new tests now.

clients/include/aux/testing_set_get_matrix.hpp
daineAMD: Created LWPMLSE-504 for bad_arg tests in hipblas.

PR #709, opened 2023-11-06
Removing some temporary changes needed for yaml testing migration

daineAMD:


clients/common/utility.cpp
```diff
@@ -218,12 +218,16 @@ hipblasLocalHandle::~hipblasLocalHandle()
...
+        if(hipStatus != hipSuccess)
+            std::cerr << "error freeing hip memory in hipblasLocalHandle: "
+                      << hipGetErrorString(hipStatus) << "\n";
     }
     hipblasStatus_t status = hipblasDestroy(m_handle);
     if(status != HIPBLAS_STATUS_SUCCESS)
     {
-        printf("hipblasDestroy error!\n");
+        std::cerr << "hipblasDestroy error: " << hipblasStatusToString(status) << "\n";
     }
```
TorreZuk: After this you might also want to do the hipGetLastError check like I added to rocblas teardown.

clients/include/d_vector.hpp
```diff
@@ -76,13 +77,15 @@ class d_vector
             if(PAD > 0)
             {
                 // Copy guard to device memory before allocated memory
-                CHECK_HIP_ERROR(hipMemcpy(d, guard, sizeof(guard), hipMemcpyHostToDevice));
+                if(hipMemcpy(d, guard, sizeof(guard), hipMemcpyHostToDevice))
+                    std::cerr << "Error: hipMemcpy pre-guard copy failure." << std::endl;
```
TorreZuk, line 81: Does this ensure a failed test?

TorreZuk (approved):
LGTM if total tests match expected count.  Points raised can be addressed later.

clients/common/utility.cpp
daineAMD: Added hipGetLastError check as suggested, but this is a hipblas failure not hip, so needed to add gtest test

clients/include/d_vector.hpp
daineAMD: With hipGetLastError added, now it does.

TorreZuk (approved):
LGTM

PR #710, opened 2023-11-07
hip platform selection using hipconfig --platform and HIP_PLATFORM

daineAMD:
Opening for initial review. Have to test with different hip versions to see if it works with hip < 5.7.
See https://github.com/ROCmSoftwarePlatform/roclibs/blob/main/rfcs/0001-hip-platform-selection.md


CMakeLists.txt
```diff
@@ -100,22 +100,29 @@ endif( )
...
 
-# Find CUDA if the user wants a CUDA version.
-option(USE_CUDA "Look for CUDA and use that as a backend if found" OFF)
-if (USE_CUDA)
-	find_package( CUDA REQUIRED )
+# Deprecated USE_CUDA option
+if(DEFINED USE_CUDA)
+  if(USE_CUDA)
+    set(ENV{HIP_PLATFORM} nvidia)
+    message(DEPRECATED "USE_CUDA is deprecated (use HIP_PLATFORM=nvidia)")
```
TorreZuk: Is it an env variable or cmake variable for HIP_PLATFORM ?  You set env but use variable so it is confusing.

daineAMD: If the user wants to override the auto-detected platform, they can set the environment variable HIP_PLATFORM. Once find_package(hip) is called, it sets the variable HIP_PLATFORM to be used in cmake. I've added a link to the document Cory created in the description.

TorreZuk: So would -DHIP_PLATFORM=nvidia be the normal usage for cuda or env if you have both installed?

TorreZuk (approved):
If builds fine by me but get Cory to approve.

CMakeLists.txt
daineAMD: If you're building with cmake I believe you'll still want to set the environment variable to force HIP to find the platform you want.

CMakeLists.txt
```diff
@@ -100,22 +100,29 @@ endif( )
...
-    # support for cuda backend with hip < 6.0
-    if( NOT hip_FOUND )
-        find_package( HIP MODULE REQUIRED )
-        list( APPEND HIP_INCLUDE_DIRS "${HIP_ROOT_DIR}/include" )
-    endif()
-else()
-    find_package( hip REQUIRED CONFIG PATHS ${HIP_DIR} ${ROCM_PATH} /opt/rocm)
+find_package( hip CONFIG PATHS ${HIP_DIR} ${ROCM_PATH} /opt/rocm )
+
+# support for cuda backend with hip < 6.0
```
cgmb, line 117: This will work for old version of HIP as long as the user doesn't have HIP for the AMD platform installed (as otherwise it might have been found but not work). I'm not sure if it's worth working around that issue. Up to you.

cgmb (approved):
LGTM. It'll be nice that platform selection 'just works' by default for users that are building with recent versions of HIP.

We'll be able to drop even more of the logic from the library build files if `hip::host` can be updated to include the necessary interface properties for the NVIDIA platform.

Thanks for bringing that proof-of-concept into a reality, @daineAMD!

PR #711, opened 2023-11-08
Temporarily remove c-example from nv. build

daineAMD:
Revert once SWDEV-431568 is resolved.

This removes our C example from being built when on nvidia platform to get CI back in a good state.


TorreZuk (approved):
LGTM

daineAMD:
Ironically cuda CI nodes seem to be down. Built manually ok on cuda and other CI seems fine so merging.

PR #712, opened 2023-11-09
General cleanup for clang warnings

daineAMD:
- Adds hipblas_abs to use instead of std::abs for accurate abs for our half/bf16/complex types.
- Adds default cases in switch statements.

This should clean up hipcc/clang++ build warnings.


TorreZuk (approved):
LGTM

PR #719, opened 2023-11-21
Temporary fix for --rocblas-path/--rocsolver-path in rmake.py

daineAMD:
This is a work in progress to get some review.
- This PR should work on linux for g++
- Need to test manually on windows yet
- Seems to me that these flags still don't work when using ./install.sh or rmake.py when using hip-clang, need to investigate

Next release I'll update install.sh/rmake.py to match rocBLAS more closely so will be further changes then.


TorreZuk (approved):
Okay but you say they don't work?

jsandham:
Copied this to hipsparse and it works there on windows and linux.

jsandham (approved):

daineAMD:
Seems like it works fine on windows for me, just not when using --hip-clang flag in conjunction with --rocblas-path flag on linux.
Might not be worth worrying about since (a) it's already been broken for ??? time and (b) we'll be moving away from hipcc anyway.

daineAMD:
Will wait for windows CI to run with recent changes just to be sure.

PR #720, opened 2023-11-21
Rename 'aux' dirs to 'auxil'

daineAMD:
Windows doesn't like files/folders named 'aux', see https://learn.microsoft.com/en-ca/windows/win32/fileio/naming-a-file?redirectedfrom=MSDN#win32-file-namespaces


TorreZuk (approved):
Okay but I think I don't like auxil either, not sure why though :-)

daineAMD:
> Okay but I think I don't like auxil either, not sure why though :-)

Any preference on another name? 'misc', 'auxiliary', 'other', 'blas_aux', 'blas0'? We don't have these folders in rocblas

TorreZuk:
> > Okay but I think I don't like auxil either, not sure why though :-)
> 
> Any preference on another name? 'misc', 'auxiliary', 'other', 'blas_aux', 'blas0'? We don't have these folders in rocblas

It is fine.   Etymology fits it

amcamd (approved):
good thing aux and auxil occur in auxiliary and not any other word I can find.

TorreZuk:
I think that it built may be enough results to merge.  The test looks like some connection failure so can just review the next develop test run

PR #723, opened 2023-11-24
Bad arg tests for L1

daineAMD:
Makes it easier to track discrepancies between backends, may or may not be worth adding for all functions.


clients/include/blas1/testing_copy.hpp
```diff
@@ -36,6 +36,31 @@ inline void testname_copy(const Arguments& arg, std::string& name)
...
+    device_vector<T> dy(N * incy);
+
+    EXPECT_HIPBLAS_STATUS(hipblasCopyFn(nullptr, N, dx, incx, dy, incy),
+                          HIPBLAS_STATUS_NOT_INITIALIZED);
+#ifndef __HIP_PLATFORM_NVCC__
+    EXPECT_HIPBLAS_STATUS(hipblasCopyFn(handle, N, nullptr, incx, dy, incy),
+                          HIPBLAS_STATUS_INVALID_VALUE);
+    EXPECT_HIPBLAS_STATUS(hipblasCopyFn(handle, N, dx, incx, nullptr, incy),
+                          HIPBLAS_STATUS_INVALID_VALUE);
+#endif
```
TorreZuk: So how are these inconsistencies tracked?   Do we want to note them with some convention?

clients/include/blas1/testing_copy_batched.hpp
```diff
@@ -36,6 +36,31 @@ inline void testname_copy_batched(const Arguments& arg, std::string& name)
...
+
+    device_batch_vector<T> dx(N, incx, batch_count);
+    device_batch_vector<T> dy(N, incy, batch_count);
+
+    EXPECT_HIPBLAS_STATUS(hipblasCopyBatchedFn(nullptr, N, dx, incx, dy, incy, batch_count),
+                          HIPBLAS_STATUS_NOT_INITIALIZED);
+    EXPECT_HIPBLAS_STATUS(hipblasCopyBatchedFn(handle, N, nullptr, incx, dy, incy, batch_count),
+                          HIPBLAS_STATUS_INVALID_VALUE);
+    EXPECT_HIPBLAS_STATUS(hipblasCopyBatchedFn(handle, N, dx, incx, nullptr, incy, batch_count),
+                          HIPBLAS_STATUS_INVALID_VALUE);
```
TorreZuk, line 61: So batched is consistent....

TorreZuk (approved):
All looks good if tests pass. 
Only question is when is difference with #ifndef __HIP_PLATFORM_NVCC__ something to fix, and make a TODO?   Assume none of the coded ones.

clients/include/blas1/testing_copy.hpp
daineAMD: Could maybe do it in the yaml files somehow, but might get messy. I'll see if I can think of something better.

clients/include/blas1/testing_copy_batched.hpp
daineAMD: cuBLAS has limited batched support, no L1s are supported so these are excluded in the yaml files from cuBLAS backend.

clients/include/blas1/testing_copy.hpp
daineAMD: You can take a look at the last commit, it adds a flag to hipblas-arguments to enable "all" bad_arg tests. Not sure if it adds value or not, but essentially could just change .yaml files if we want to add these tests to cublas backend.

clients/include/hipblas_common.yaml
```diff
@@ -594,6 +595,7 @@ Defaults:
   backend_flags: ALL_BACKEND
   api: C
   initialization: rand_int
+  bad_arg_all: true
   compute_type_gemm: 2
```
TorreZuk, line 599: Okay so bad_arg_all is just to let you test the expected failures for inconsistent backend?   Guess it make sense, but a || backend_all could be another option.   This is just likely applicable to bad_arg I assume

daineAMD: Yep, that's right. I'll leave it for now and see how the rest of L2/L3 bad args looks.

PR #725, opened 2023-11-27
install data files for hipblas-test

daineAMD:
Will go into 6.1 as well.


TorreZuk (approved):
Matches rocblas so should work but you'll need to manually test as no CI coverage

PR #729, opened 2023-11-29
Add bad-arg tests for all L2

daineAMD:
Will fail a couple nvidia backend tests until #728 goes through

### DIFF
--- a/clients/gtest/blas2/ger_gtest.yaml
+++ b/clients/gtest/blas2/ger_gtest.yaml
@@ -63,7 +63,6 @@ Tests:
       - ger_bad_arg: *single_double_precisions
       - geru_bad_arg: *single_double_precisions_complex
       - gerc_bad_arg: *single_double_precisions_complex
-    precision: *single_double_precisions_complex_real
     api: [ FORTRAN, C ]
     backend_flags: AMD
 
@@ -73,8 +72,6 @@ Tests:
       - ger_bad_arg: *single_double_precisions
       - geru_bad_arg: *single_double_precisions_complex
       - gerc_bad_arg: *single_double_precisions_complex
-    precision: *single_double_precisions_complex_real
-    precision: *single_double_precisions_complex_real
     api: [ FORTRAN, C ]
     bad_arg_all: false
     backend_flags: NVIDIA
--- a/clients/gtest/blas2/ger_gtest.yaml
+++ b/clients/gtest/blas2/ger_gtest.yaml
@@ -56,4 +56,38 @@ Tests:
     stride_scale: [ 1.0, 2.5 ]
     api: [ FORTRAN, C ]
     backend_flags: AMD
+
+  - name: ger_bad_arg
+    category: pre_checkin
+    function:
+      - ger_bad_arg: *single_double_precisions
+      - geru_bad_arg: *single_double_precisions_complex
+      - gerc_bad_arg: *single_double_precisions_complex
+    precision: *single_double_precisions_complex_real
+    api: [ FORTRAN, C ]
+    backend_flags: AMD
+
+  - name: ger_bad_arg
+    category: pre_checkin
+    function:
+      - ger_bad_arg: *single_double_precisions
+      - geru_bad_arg: *single_double_precisions_complex
+      - gerc_bad_arg: *single_double_precisions_complex
+    precision: *single_double_precisions_complex_real
+    precision: *single_double_precisions_complex_real
+    api: [ FORTRAN, C ]
+    bad_arg_all: false
+    backend_flags: NVIDIA
+
+  - name: ger_bad_arg
+    category: pre_checkin
+    function:
+      - ger_batched_bad_arg: *single_double_precisions
+      - ger_strided_batched_bad_arg: *single_double_precisions
+      - geru_batched_bad_arg: *single_double_precisions_complex
+      - geru_strided_batched_bad_arg: *single_double_precisions_complex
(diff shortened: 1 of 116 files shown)


TorreZuk:
I'll review this one now

clients/gtest/blas2/ger_gtest.yaml
```diff
@@ -56,4 +56,38 @@ Tests:
...
+    backend_flags: AMD
+
+  - name: ger_bad_arg
+    category: pre_checkin
+    function:
+      - ger_bad_arg: *single_double_precisions
+      - geru_bad_arg: *single_double_precisions_complex
+      - gerc_bad_arg: *single_double_precisions_complex
+    precision: *single_double_precisions_complex_real
+    precision: *single_double_precisions_complex_real
```
TorreZuk: This notation seems wrong

clients/gtest/blas2/ger_gtest.yaml
```diff
@@ -56,4 +56,38 @@ Tests:
...
     api: [ FORTRAN, C ]
     backend_flags: AMD
+
+  - name: ger_bad_arg
+    category: pre_checkin
+    function:
+      - ger_bad_arg: *single_double_precisions
+      - geru_bad_arg: *single_double_precisions_complex
+      - gerc_bad_arg: *single_double_precisions_complex
+    precision: *single_double_precisions_complex_real
```
TorreZuk: dual notation not clear which is used

clients/gtest/blas2/her2_gtest.cpp
```diff
@@ -112,10 +112,16 @@ namespace
...
+                testing_her2_bad_arg<T>(arg);
             else if(!strcmp(arg.function, "her2_batched"))
                 testing_her2_batched<T>(arg);
+            else if(!strcmp(arg.function, "her2_batched_bad_arg"))
+                testing_her2_batched_bad_arg<T>(arg);
             else if(!strcmp(arg.function, "her2_strided_batched"))
                 testing_her2_strided_batched<T>(arg);
+            else if(!strcmp(arg.function, "her2_strided_batched_bad_arg"))
+                testing_her2_strided_batched_bad_arg<T>(arg);
             else
```
TorreZuk, line 125: Maybe a macro could have been of use.....

clients/include/blas2/testing_hpr.hpp
```diff
@@ -37,6 +37,68 @@ inline void testname_hpr(const Arguments& arg, std::string& name)
...
+    using U           = real_t<T>;
+    bool FORTRAN      = arg.api == hipblas_client_api::FORTRAN;
+    auto hipblasHprFn = FORTRAN ? hipblasHpr<T, U, true> : hipblasHpr<T, U, false>;
+
+    for(auto pointer_mode : {HIPBLAS_POINTER_MODE_HOST, HIPBLAS_POINTER_MODE_DEVICE})
+    {
+        hipblasLocalHandle handle(arg);
+        CHECK_HIPBLAS_ERROR(hipblasSetPointerMode(handle, pointer_mode));
+
+        hipblasFillMode_t uplo   = HIPBLAS_FILL_MODE_UPPER;
```
TorreZuk, line 52: The yaml seems to specify these but you don't use them.  I would either use or remove from yaml?
Not just this function.... but won't repeat

clients/include/blas2/testing_hpr2.hpp
```diff
@@ -37,6 +37,77 @@ inline void testname_hpr2(const Arguments& arg, std::string& name)
...
+    for(auto pointer_mode : {HIPBLAS_POINTER_MODE_HOST, HIPBLAS_POINTER_MODE_DEVICE})
+    {
+        hipblasLocalHandle handle(arg);
+        CHECK_HIPBLAS_ERROR(hipblasSetPointerMode(handle, pointer_mode));
+
+        hipblasFillMode_t uplo   = HIPBLAS_FILL_MODE_UPPER;
+        int64_t           N      = 100;
+        int64_t           incx   = 1;
+        int64_t           incy   = 1;
+        int64_t           A_size = int64_t(N) * (N + 1) / 2;
```
TorreZuk: Fine but you could remove the cast if all moving to int64_t which is good.

TorreZuk (approved):
Wow a lot of additions!     Most comments I made are minor.

clients/include/blas2/testing_hpr.hpp
daineAMD: Do you mind taking a look at 3a0fc1c to see if you like this change better? Tests needed these params for test name generation before.

TorreZuk: Well that is fine too, do you ever see potential in testing an invalid enum with bad arg?  If you just use arg.uplo instead of a constant this could allow it.  Only concern was putting something in the yaml you expect it may be used, so that fact that it isn't may just lead poor understanding of the test set, or test coverage.

daineAMD: I don't think I see an advantage using yaml for things like enums; invalid enums are just tested by coding it in the testing_*.hpp file right now which I think is better. Could just take the regular enum from the yaml file like you say not sure there's much of an advantage.

Not just the enums but things like sizes in the yaml just don't make much sense for the bad_arg tests which is why I made the change. Seeing a test name like `gemv_bad_arg_a_type_f32_r` makes more sense than having a bunch of params that don't really apply to that test.

TorreZuk: Good I see they get removed from the yaml

PR #733, opened 2023-12-06
Enabling and adding bad-arg tests for rocsolver

daineAMD:
Should be the final PR in the bad-arg saga.

There already existed some bad-arg tests for solver, but weren't enabled in the yaml files.

Note some strange behaviour with cuBLAS backend for gels, doesn't seem to make sense for me so left those tests out for now.

### DIFF
--- a/clients/gtest/solver/getrf_gtest.cpp
+++ b/clients/gtest/solver/getrf_gtest.cpp
@@ -73,17 +73,22 @@ namespace
             switch(GETRF_TYPE)
             {
             case GETRF:
-                return !strcmp(arg.function, "getrf");
+                return !strcmp(arg.function, "getrf") || !strcmp(arg.function, "getrf_bad_arg");
             case GETRF_BATCHED:
-                return !strcmp(arg.function, "getrf_batched");
+                return !strcmp(arg.function, "getrf_batched")
+                       || !strcmp(arg.function, "getrf_batched_bad_arg");
             case GETRF_STRIDED_BATCHED:
-                return !strcmp(arg.function, "getrf_strided_batched");
+                return !strcmp(arg.function, "getrf_strided_batched")
+                       || !strcmp(arg.function, "getrf_strided_batched_bad_arg");
             case GETRF_NPVT:
-                return !strcmp(arg.function, "getrf_npvt");
+                return !strcmp(arg.function, "getrf_npvt")
+                       || !strcmp(arg.function, "getrf_npvt_bad_arg");
             case GETRF_NPVT_BATCHED:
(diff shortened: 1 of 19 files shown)


clients/gtest/solver/getrf_gtest.cpp
```diff
@@ -73,17 +73,22 @@ namespace
...
+                return !strcmp(arg.function, "getrf_npvt")
+                       || !strcmp(arg.function, "getrf_npvt_bad_arg");
             case GETRF_NPVT_BATCHED:
-                return !strcmp(arg.function, "getrf_npvt_batched");
+                return !strcmp(arg.function, "getrf_npvt_batched")
+                       || !strcmp(arg.function, "getrf_npvt_batched_bad_arg");
             case GETRF_NPVT_STRIDED_BATCHED:
-                return !strcmp(arg.function, "getrf_npvt_strided_batched");
+                return !strcmp(arg.function, "getrf_npvt_strided_batched")
+                       || !strcmp(arg.function, "getrf_npvt_strided_batched_bad_arg");
```
TorreZuk, line 91: Curious what the assembly is on all these, it isn't constexpr so can the switch be optimized out?

clients/include/solver/testing_gels_batched.hpp
```diff
@@ -94,57 +91,45 @@ void testing_gels_batched_bad_arg(const Arguments& arg)
...
     EXPECT_HIPBLAS_STATUS(
         hipblasGelsBatchedFn(handle, opN, M, N, nrhs, dAp, lda, dBp, ldb, &info, dInfo, -1),
         HIPBLAS_STATUS_INVALID_VALUE);
-    expectedInfo = -11;
-    unit_check_general(1, 1, 1, &expectedInfo, &info);
+
+    if(arg.bad_arg_all)
+    {
+        expectedInfo = -11; // cublas gets -8
+        unit_check_general(1, 1, 1, &expectedInfo, &info);
```
TorreZuk, line 131: If you know the expected outputs for both you don't want to switch on backend?

TorreZuk (approved):
LGTM

clients/include/solver/testing_gels_batched.hpp
daineAMD: Yeah could, I didn't because the cuBLAS results don't seem logical to me. Didn't want to assert something that I think might be a bug. But if someone can make sense of the results to me then I can make the change!

clients/gtest/solver/getrf_gtest.cpp
daineAMD: Can't remove the switch, get duplicated test names. Can make it a constexpr if statement, though.

PR #734, opened 2023-12-06
Adding conversion for fill/uplo when missing, with amd backend

daineAMD:
Changing from simple cast to use converter


TorreZuk (approved):
LGTM if test pass, what does hipFillToHCCFill stand for?

daineAMD:
Yeah should be `hipFillToRocblasFill` or `hipblasFillToRocblasFill` or something, with cuBLAS backend it's `hipFillToCudaFill`. All the conversion functions use HCC for some reason.

PR #735, opened 2023-12-07
Use arg.api instead of arg.fortran

daineAMD:
Lots of hipblas updates today.


TorreZuk (approved):
LGTM, I guess the _64 can be added to follow the BLAS style?

PR #739, opened 2023-12-12
Client side template specialization using macros

daineAMD:
Changing the client-side template specialization for calls to hipblas so less repeated code when adding 64-bit interface. Only done for copy and scal for now, will change all functions in next PR if we want this change.

Similar to rocBLAS, notable change being I've added an intermediary "Cast" layer for HIPBLAS_V2 since we need to cast from hipblasComplex (which is used in client) to hipComplex in hipblas.h

### DIFF
--- a/clients/include/hipblas.hpp
+++ b/clients/include/hipblas.hpp
@@ -50,94 +50,94 @@
 #if !defined(HIPBLAS_V2) \
     && !defined(         \
         WIN32) // HIPBLAS_V2 doesn't have fortran tests during transition period, WIN doesn't have fortran tests
-#define MAP2CF3(FN, A, PFN)         \
-    template <>                     \
-    static auto FN<A, false> = PFN; \
-    template <>                     \
-    static auto FN<A, true> = PFN##Fortran
-#define MAP2CF4(FN, A, B, PFN)         \
-    template <>                        \
-    static auto FN<A, B, false> = PFN; \
-    template <>                        \
-    static auto FN<A, B, true> = PFN##Fortran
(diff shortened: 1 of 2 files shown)


clients/include/hipblas.hpp
```diff
@@ -28,59 +28,321 @@
...
+                                                 int             incx,
+                                                 hipblasStride   stridex,
+                                                 int             batch_count);
+
+    MAP2CF(hipblasScal, float, float, hipblasSscal);
+    MAP2CF(hipblasScal, double, double, hipblasDscal);
+    MAP2CF_V2(hipblasScal, hipblasComplex, hipblasComplex, hipblasCscal);
+    MAP2CF_V2(hipblasScal, hipblasDoubleComplex, hipblasDoubleComplex, hipblasZscal);
+    MAP2CF_V2(hipblasScal, hipblasComplex, float, hipblasCsscal);
+    MAP2CF_V2(hipblasScal, hipblasDoubleComplex, double, hipblasZdscal);
```
TorreZuk, line 282: Okay so this will allow V2 with same function pointer helper?

TorreZuk (approved):
Fine by me if it simplifies.  Tough to review but as long as test names all appear as expected macro magic must be working

clients/include/hipblas.hpp
daineAMD: Yep that's the idea. The _V2 version should just add "Cast" to the end of the hipblas API, e.g. hiblasZdScal(Cast). Only to be used with complex functions.

The macros are copied directly from rocBLAS, should only have this addition.

PR #740, opened 2023-12-13
Client side template specialization using macros for all functions

daineAMD:
Remaining changes following #739
Huge changes, sorry. Changes follow the same pattern from #739 and just applies it to all functions.


TorreZuk (approved):
LGTM diffs were confused by github, looked at file but not diff.  can you confirm test counts didn't change.

daineAMD:
Yep, local build of ~develop has same amount of tests for hipblas-test and hipblas_v2-test as CI.

PR #741, opened 2023-12-13
Allow debug builds in jenkins

daineAMD:


TorreZuk (approved):
I was just thinking of adding this exact change :-)

TorreZuk:
I can make debug packages locally, maybe look at rocblas groovy for hints.   I would think maybe we should remove the debug suffix unless you heard otherwise?  We don't use them on rocblas do we?  
DEBUG_POSTFIX I am referring to as suffix

daineAMD:
We don't seem to have them in rocblas as far as I can tell. I don't think anyone tests our debug builds as far as I know.
I've removed DEBUG_POSTFIX in my last commit. Maybe it should be it's own PR though, let me know what you think.

TorreZuk:
Well the testing can run the same executable so fine to leave it in the POSTFIX change in here.  If someone needed it back it is a couple lines.

PR #742, opened 2023-12-13
Adding int32 interfaces for near_check_general

daineAMD:
and removing unneeded include for samples


TorreZuk (approved):
So why would only debug look for these symbols?    Include change makes sense

TorreZuk:
Thanks, This fixed my local debug build so guess you can merge and see if your other PR the re-run passes.  Should try and figure out the int function reference variation later....

daineAMD:
type_dispatch.hpp has a call to testing_gemm_ex with int32_t datatype which calls near_check. Not sure what's with the difference between release and debug builds though. Merging to try #741 again.

TorreZuk:
Okay  @daineAMD here is my guess, only when inlining can it actually figure out that the test is not invoked, thus the debug build requires the symbols in case it is called thus the reference is inserted.

PR #744, opened 2023-12-14
64-bit infrastructure and copy implementation

daineAMD:
Should wait for reference code, but pretty much complete otherwise


TorreZuk (approved):
Okay that looks like it doesn't overlap so LGTM if you get it passing.
I can push my WIP this afternoon as got distracted by a triage

PR #745, opened 2023-12-14
ILP64 reference with helpers 64bit updates

TorreZuk:
* WIP will crash in tests
* Adds OpenMP and AOCL for ILP64 reference


clients/common/norm.cpp
```diff
@@ -44,122 +46,135 @@
...
+//             int*                  incx,
+//             hipblasDoubleComplex* y,
+//             int*                  incy);
 
 #ifdef __cplusplus
 }
 #endif
 
+template <typename T>
+void m_axpy_64(int64_t N, T* alpha, T* x, int64_t incx, T* y, int64_t incy)
```
daineAMD, line 76: Why do we need this one here?

clients/include/hipblas_vector.hpp
```diff
@@ -516,13 +533,13 @@ void hipblas_init_alternating_template(U& that, T rand_gen(), T rand_gen_alt(),
 
     for(int b = 0; b < that.batch_count(); ++b)
```
daineAMD: should be int64_t here

clients/common/norm.cpp
TorreZuk: What do you mean exactly, why not reuse ref_  form?   It alllows a self contained 64bit support in norm.cpp.  Windows will get additions to allow 32bit but I think linux may be stuck with always 64bit reference.   Due to variations it maybe only solver -> LAPACKE will be wrapped with #ifdef WIN32 to allow 32bit API visibility.

daineAMD (approved):
Cuda pipeline can't find omp looks like.

TorreZuk:
> Cuda pipeline can't find omp looks like.

Okay thanks, I'll probably have to experiment with the docker.  No shortage of problems to work around.

TorreZuk:
Okay @daineAMD if you can re-review as a lot has changed but this is working on solver now with libflame for lapack.  So is windows.  Unfortunately seems like cuda pipline is stopped were you already inquiring on this or should I ?

daineAMD (approved):
Great to have the matrix init code, LGTM I think.

PR #749, opened 2023-12-19
Adding gemvBatched and gemvStridedBatched with cuda backend

daineAMD:
This adds hipblasXgemvBatched and hipblasXgemvStridedBatched when using cuBLAS 11.7.0 or greater as a backend.


clients/include/blas2/testing_gemv_batched.hpp
```diff
@@ -311,6 +316,7 @@ void testing_gemv_batched(const Arguments& arg)
                                                       batch_count);
         EXPECT_HIPBLAS_STATUS(
             actual, (invalid_size ? HIPBLAS_STATUS_INVALID_VALUE : HIPBLAS_STATUS_SUCCESS));
+	*/
         return;
     }
```
TorreZuk, line 324: Why not test with guard?   I would put in a TODO, but if you don't want double coverage put in a SKIPPED so these can be removed when seen in test run.

library/src/nvidia_detail/hipblas.cpp
```diff
@@ -4847,8 +4876,29 @@ hipblasStatus_t hipblasZgemvBatched(hipblasHandle_t                   handle,
                                     hipblasDoubleComplex* const       y[],
                                     int                               incy,
                                     int                               batchCount)
+try
 {
+#if CUBLAS_VERSION >= 110700
+    return hipCUBLASStatusToHIPStatus(cublasZgemvBatched((cublasHandle_t)handle,
```
TorreZuk, line 4882: Probably should make a future TODO ticket to review moving to a getProcAddress query system so we can cover a few major releases with the same build.   These are for the C API so shouldn't be too tricky.

TorreZuk (approved):
LGTM future work suggestions

library/src/nvidia_detail/hipblas.cpp
daineAMD: Created [LWPMLSE-523](https://ontrack-internal.amd.com/browse/LWPMLSE-523)

PR #754, opened 2023-12-21
Asum 64-bit interface

daineAMD:
Is it easiest to review these 1-at-a-time like this? Otherwise I can put a handful of functions in each PR.


library/src/nvidia_detail/hipblas.cpp
```diff
@@ -1020,6 +1020,101 @@ catch(...)
...
+
+hipblasStatus_t hipblasDasum_64(
+    hipblasHandle_t handle, int64_t n, const double* x, int64_t incx, double* result)
+try
+{
+#if CUBLAS_VER_MAJOR >= 12
+    return hipCUBLASStatusToHIPStatus(cublasDasum_64((cublasHandle_t)handle, n, x, incx, result));
+#else
+    return HIPBLAS_STATUS_NOT_SUPPORTED;
+#endif
```
TorreZuk, line 1047: It might help to put these 5 lines in a macro.

TorreZuk (approved):
LGTM Whatever number of functions is easiest is fine.  If you have python that helps then more is fine.

library/src/nvidia_detail/hipblas.cpp
daineAMD: Ah yeah, I'll do that in the next one if I remember over the break :smile:

PR #756, opened 2023-12-22
Dot 64-bit interface

daineAMD:


TorreZuk (approved):
LGTM, not sure if you have local build windows to fix the warnings for dangling else, they were on earlier _64 so not sure if it is one offs or in the pattern of this PR as well.  Minor.

PR #760, opened 2024-01-04
Rot* 64-bit interface

daineAMD:
- so far no "real" 64-bit tests, I'll open a test PR with some added first.
- this adds 64-bit implementation of rot, rotm, rotg, and rotmg
- cuBLAS documentation seems to be wrong here, says it supports 64-bit implementation of rotg and rotmg, but cublas_v2.h has lines like `#define cublasSrotg_64 cublasSrotg_64_v2` but `cublasSrotg_64_v2` is never defined in cublas_api.h; so hipBLAS isn't supporting this with cuBLAS backend.


clients/common/hipblas_template_specialization.cpp
```diff
@@ -1212,6 +1212,62 @@ hipblasStatus_t hipblasZdrotCast(hipblasHandle_t       handle,
...
+                                   int64_t               incx,
+                                   hipblasComplex*       y,
+                                   int64_t               incy,
+                                   const float*          c,
+                                   const hipblasComplex* s)
+{
+    return hipblasCrot_64(
+        handle, n, (hipComplex*)x, incx, (hipComplex*)y, incy, c, (const hipComplex*)s);
+}
+
```
TorreZuk, line 1372: Wonder if you should split this into L1, L2, L3, solver files?  Not sure if is one of the slower ones or not....

clients/include/blas1/testing_rotm_batched.hpp
```diff
@@ -118,9 +122,10 @@ void testing_rotm_batched(const Arguments& arg)
     hipblas_init_vector(hy, arg, hipblas_client_alpha_sets_nan, false);
     hipblas_init_vector(hdata, arg, hipblas_client_alpha_sets_nan, false);
 
-    for(int b = 0; b < batch_count; b++)
+    for(int64_t b = 0; b < batch_count; b++)
     {
-        ref_rotmg<T>(&hdata[b][0], &hdata[b][1], &hdata[b][2], &hdata[b][3], hparam[b]);
+        int b2 = b;
+        ref_rotmg<T>(&hdata[b2][0], &hdata[b2][1], &hdata[b2][2], &hdata[b2][3], hparam[b2]);
```
TorreZuk, line 128: Not sure why you do this int b2, maybe you haven't rebased.  If it is missing you can add another operator[](int64_t) I am pretty sure I added to something just can't recall what.

library/src/nvidia_detail/hipblas.cpp
```diff
@@ -4605,6 +5441,35 @@ catch(...)
...
+    hipblasHandle_t handle, float* d1, float* d2, float* x1, const float* y1, float* param)
+try
+{
+    // cuBLAS documentation claims support here but doesn't seem to have it in cuda 12.3
+    // #if CUBLAS_VER_MAJOR >= 12
+    //     return hipCUBLASStatusToHIPStatus(
+    //         cublasSrotmg_64((cublasHandle_t)handle, d1, d2, x1, y1, param));
+    // #else
+    return HIPBLAS_STATUS_NOT_SUPPORTED;
+    // #endif
```
TorreZuk: Well you could do what we do internally and just call their non _64 function.  We just added it for a more consistent  API.

TorreZuk (approved):
LGTM if all passing.  Documentation form for all and testing enablement can come later.

clients/include/blas1/testing_rotm_batched.hpp
daineAMD: Yeah it's probably been fixed already. I'll change it all back when I enable full testing.

library/src/nvidia_detail/hipblas.cpp
daineAMD: Ah yeah true I'll do that.

clients/common/hipblas_template_specialization.cpp
daineAMD: I can play around with this once everything is in.

PR #761, opened 2024-01-05
Add gpu_arch to filter and enable 64-bit asum tests

daineAMD:


TorreZuk (approved):
Looks good if expected tests run

PR #762, opened 2024-01-05
Scal and swap 64-bit interface

daineAMD:


TorreZuk (approved):
LGTM, hopefully with the rebase you'll get all passing now

TorreZuk:
@daineAMD I think the cuda is due to Eiden changing the docker again build to #13279 , as my fix PR all passed cuda before I merged :-)

TorreZuk:
http://math-ci.amd.com/blue/organizations/jenkins/main%2Fprecheckin-cuda%2FhipBLAS/detail/PR-759/2/pipeline

.jenkins/precheckin-cuda.groovy
```diff
@@ -89,6 +89,6 @@ ci: {
             properties(auxiliary.addCommonProperties(property))
     }
 
-    String hostBuildCommand = './install.sh -c --compiler=g++ --cuda'
+    String hostBuildCommand = 'HIP_PLATFORM=nvidia ./install.sh -c --compiler=g++'
```
TorreZuk: Okay so this is wrong @daineAMD.  Nothing in install.sh reads that env variable to set build configuration.  
I would suggest this needs to be added in separate PR.  Also the move from install.sh code duplication of rmake.py should be addressed soon.  I was trying to explain whether CI failures which seem to be using wrong reference library were due to CI pipeline bug or stale cache but this should be reverted first to see if it passes.  Then I can make fix PR or you can, we can discuss.  Same goes for moving build code from install into rmake I can look if you want.

daineAMD: Yep I'll remove that now. I do have a ticket for fixing install.sh/rmake.py duplication, may or may not get to it this release.

daineAMD:
Windows CI failed but passed previous build so fine to merge I think.

PR #768, opened 2024-01-13
Enable all L1 64-bit tests

daineAMD:
Enabling all _64 tests. Only run on gfx90a, not tested with cublas backend right now.

### DIFF
--- a/clients/include/unit.h
+++ b/clients/include/unit.h
@@ -83,9 +83,9 @@ void unit_check_error(T error, T tolerance)
 template <typename T, typename Tex = T>
 void unit_check_nrm2(T cpu_result, T gpu_result, int64_t vector_length)
 {
-    T allowable_error = vector_length * hipblas_type_epsilon<T> * cpu_result;
+    T allowable_error = vector_length * hipblas_type_epsilon<Tex> * cpu_result;
     if(allowable_error == 0)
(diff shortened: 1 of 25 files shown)


clients/include/unit.h
```diff
@@ -81,27 +81,27 @@ void unit_check_error(T error, T tolerance)
 }
 
 template <typename T, typename Tex = T>
-void unit_check_nrm2(T cpu_result, T gpu_result, int vector_length)
+void unit_check_nrm2(T cpu_result, T gpu_result, int64_t vector_length)
 {
-    T allowable_error = vector_length * std::numeric_limits<Tex>::epsilon() * cpu_result;
+    T allowable_error = vector_length * hipblas_type_epsilon<T> * cpu_result;
     if(allowable_error == 0)
-        allowable_error = vector_length * std::numeric_limits<Tex>::epsilon();
```
TorreZuk, line 88: Was Tex not valid here?

TorreZuk (approved):
Hopefully all pass

daineAMD:
All passed now. Note that test time on gfx90a went from (49min + 32min) to (74min + 59min), taking ~50 minutes longer now if running both hipblas-test and hipblas_v2-test. Should go through test suite again and remove unnecessary tests, start using the categories in CI, and can communicate with downstream testing as well.

PR #770, opened 2024-01-15
Add documentation for 64-bit interface

daineAMD:
Copied documentation from rocBLAS minus the comment about performance since we hipBLAS doesn't have any bearing on performance.


TorreZuk (approved):
LGTM, Note was pushed up to the rst doc so the hyperlinks could work in rocBLAS so I assume the same work be true here.  Would be placed in functions.rst only after each function name you could put the sentence with hyperlink up to explanation of ILP64 Interface.  Either way okay

PR #771, opened 2024-01-15
Change API from *_64_v2 to *_v2_64

daineAMD:
Find and replace "_64_v2" to "_v2_64". This is what cuBLAS does and it makes more sense to me to have _64 as the suffix, also simplifies documentation.


TorreZuk (approved):
Fine by me

PR #773, opened 2024-01-15
Allocated batched test memory in contiguous memory

daineAMD:
Use contiguous memory when allocating batched vectors for test code, similar to rocBLAS. Previously we were allocating a bunch of memory for padding for each batch, which isn't scalable to the large-batch 64-bit tests.

This should solve the failures in #768, or I can just remove/reduce the batched testing since it isn't necessarily hipBLAS' job to test batch sizes larger than the block size.


TorreZuk (approved):
LGTM doesn't hurt to match rocBLAS test pattern.

PR #785, opened 2024-02-07
cmake: add rocblas/rocsolver as package deps if building static lib

evetsso:
When a static library is built, no linking is done.  (Basically static libraries are just archives of object files.)

So if `libfoo.a` depends on `libbar.a`, CMake just has to remember that when you link `libfoo.a`, it also needs to pull in `libbar.a`.  The problem is that CMake only knows about `libbar.a` if a `find_package(bar)` is done.  This PR fixes our generated `foo-config.cmake` file to find this dependency, when we're building a static library.

### DIFF
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -64,6 +64,8 @@ add_library( hipblas
 )
 add_library( roc::hipblas ALIAS hipblas )
 
+set(static_depends)
+
 # Build hipblas from source on AMD platform
 if(HIP_PLATFORM STREQUAL amd)
   if( NOT TARGET rocblas )
@@ -76,6 +78,7 @@ if(HIP_PLATFORM STREQUAL amd)
     else()
       find_package( rocblas REQUIRED CONFIG PATHS /opt/rocm /opt/rocm/rocblas )
     endif( )
+    list(APPEND static_depends PACKAGE rocblas)
   endif( )
 
   target_link_libraries( hipblas PRIVATE roc::rocblas hip::host )
@@ -98,6 +101,7 @@ if(HIP_PLATFORM STREQUAL amd)
         find_package( rocsolver REQUIRED CONFIG PATHS /opt/rocm /opt/rocm/rocsolver /usr/local/rocsolver )
       endif()
     endif( )
+    list(APPEND static_depends PACKAGE rocsolver)
     target_link_libraries( hipblas PRIVATE roc::rocsolver )
   endif( )
 
@@ -180,6 +184,7 @@ if(HIP_PLATFORM STREQUAL amd)
     rocm_export_targets(
         TARGETS roc::hipblas
 	DEPENDS PACKAGE hip
+	STATIC_DEPENDS ${static_depends}
 	NAMESPACE roc::
     )
 else( )
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -78,9 +78,9 @@ if(HIP_PLATFORM STREQUAL amd)
     else()
       find_package( rocblas REQUIRED CONFIG PATHS /opt/rocm /opt/rocm/rocblas )
     endif( )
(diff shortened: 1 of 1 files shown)


library/src/CMakeLists.txt
```diff
@@ -180,6 +184,7 @@ if(HIP_PLATFORM STREQUAL amd)
     rocm_export_targets(
         TARGETS roc::hipblas
 	DEPENDS PACKAGE hip
+	STATIC_DEPENDS ${static_depends}
 	NAMESPACE roc::
```
TorreZuk, line 188: I guess this is empty list for CUDA backend?   Should probably add cuda static_depends as they already provide static libs so we can test already.  These changes aren't tested in CI so can you post test results @evetsso ?  Thanks.

library/src/CMakeLists.txt
```diff
@@ -76,6 +78,7 @@ if(HIP_PLATFORM STREQUAL amd)
     else()
       find_package( rocblas REQUIRED CONFIG PATHS /opt/rocm /opt/rocm/rocblas )
     endif( )
+    list(APPEND static_depends PACKAGE rocblas)
   endif( )
 
   target_link_libraries( hipblas PRIVATE roc::rocblas hip::host )
```
TorreZuk: Probably in wrong scope, I would suggest move it to right above target_link_libraries line for consistency.
Will need hip::host as well in this list.  
 But are namespace targets not allowed ?  Could create rocm_link_libraries that adds to both?

TorreZuk (commented):
We only use STATIC_DEPENDS in rocblas for static builds but expect rocm_export_targets ignores so simpler if can always add as you do here.

library/src/CMakeLists.txt
evetsso: > Probably in wrong scope, I would suggest move it to right above target_link_libraries line for consistency.

The things that go into `static_depends` are package names.  I believe this is in the correct scope, because all of the logic that would find the `rocblas` package is contained in this scope.  I'm not even sure when `TARGET rocblas` might ever be true, but if it was, then `rocblas` would not be a separate package that needs to go into `static_depends`.

> Will need hip::host as well in this list.
> But are namespace targets not allowed ? Could create rocm_link_libraries that adds to both?

Again, these are packages, not libraries or targets.  Since `hip` is already added to non-static `DEPENDS`, and `DEPENDS` is a stronger linkage than `STATIC_DEPENDS`, we should already be doing what you're looking for.

TorreZuk: I meant inside the if( NOT TARGET rocblas ) scope, not referring here to the namespace scope.  Any thing that finds it earlier would set the target, don't think we should assume you are the first location to find something or too brittle.    So if rocblas was already found that isn't relevant it is a dependency if you link, right?  

maybe it should be in DEPENDS like rocSOLVER does, as if I understand you correctly,  static depends is only for libs not in DEPENDS.   dynamic_depends could be created for solver, and go in DEPENDS section.

evetsso: Moved the scope of the APPEND as suggested.  Dynamic depends aren't necessary, as downstream CMake usage of hipBLAS does not require a `find_package` for any of these libraries.

library/src/CMakeLists.txt
evetsso: CUDA doesn't provide CMake infrastructure for its libraries.  I don't think there's much we can do on our end, since there is nothing we could add to either DEPENDS or STATIC_DEPENDS to make anything work better.

As for CI, I've run the hipBLAS regression tests manually with all-static-math-libs and (at least excluding the stress tests) they all pass.  Hopefully we can start to get static libs going on math-ci soon to help make sure this continues to work.

TorreZuk (approved):
LGTM, @daineAMD any concerns?

daineAMD (approved):
Nope, LGTM.

PR #788, opened 2024-02-15
Invoke rmake.py when using install.sh to build

daineAMD:
- install.sh now invokes rmake.py to build
- fixes debug build on windows (copying hipblas library w/o `-d` suffix)
- builds against local gtest (in build/deps/deps-install) if possible
- removes build options --rm-legacy-include-dir, --cmakepp, and --rocm-dev build options
- print hipblas git hash during build

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -172,23 +172,6 @@ install_packages( )
     # Ideally, this could be cuda-cublas-dev, but the package name has a version number in it
     library_dependencies_ubuntu+=( "" ) # removed, use --installcuda option to install cuda
   else
-    # Custom rocm-dev installation
-    if [[ -z ${custom_rocm_dev+foo} ]]; then
-      # Install base rocm-dev package unless -v/--rocm-dev flag is passed
-      library_dependencies_ubuntu+=( "rocm-dev" )
-      library_dependencies_centos_rhel+=( "rocm-dev" )
-      library_dependencies_centos_rhel_8=( "rocm-dev" )
-      library_dependencies_fedora+=( "rocm-dev" )
-      library_dependencies_sles+=( "rocm-dev" )
-    else
-      # Install rocm-specific rocm-dev package
-      library_dependencies_ubuntu+=( "${custom_rocm_dev}" )
-      library_dependencies_centos_rhel+=( "${custom_rocm_dev}" )
-      library_dependencies_centos_rhel_8+=( "${custom_rocm_dev}" )
-      library_dependencies_fedora+=( "${custom_rocm_dev}" )
-      library_dependencies_sles+=( "${custom_rocm_dev}" )
-    fi
-
     # Custom rocblas installation
     # Do not install rocblas if --rocblas_path flag is set,
     # as we will be building against our own rocblas instead.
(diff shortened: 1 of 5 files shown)


install.sh
```diff
@@ -420,6 +336,69 @@ fi
 # The following function exits script if an unsupported distro is detected
 supported_distro
 
+function display_help()
+{
+cat <<EOF
+
+  hipBLAS library build & installation helper script.
```
TorreZuk, line 343: Okay but this is still duplicating rmake.py help right?   I think I only left in the install handled options in rocBLAS

rmake.py
```diff
@@ -59,12 +56,36 @@ def parse_args():
     parser.add_argument(      '--codecoverage', required=False, default=False, action='store_true',
                         help='Code coverage build. Requires Debug (-g|--debug) or RelWithDebInfo mode (-k|--relwithdebinfo), (optional, default: False)')
 
+    parser.add_argument(      '--compiler', type=str, required=False, default='g++', dest='compiler',
+                        help='Spcify path to host compiler.')
+
+    parser.add_argument('--cuda', '--use-cuda', dest='use_cuda', required=False, default=False,
+                        help='[DEPRECATED] Build library for CUDA backend. Deprecated, use HIP_PLATFORM environment variable to override default which is determined by `hipconfig --platform`')
```
TorreZuk: Needs action=store_true maybe or is that default action?     See cuda failed maybe something else

install.sh
```diff
@@ -453,7 +431,7 @@ declare -a cmake_client_options
 # check if we have a modern version of getopt that can handle whitespace and long parameters
 getopt -T
 if [[ $? -eq 4 ]]; then
-  GETOPT_PARSE=$(getopt --name "${0}" --longoptions help,install,codecoverage,clients,no-solver,dependencies,debug,hip-clang,no-hip-clang,compiler:,cmake_install,cuda,use-cuda,cudapath:,installcuda,installcudaversion:,static,cmakepp,relocatable:,rocm-dev:,rocblas:,rocblas-path:,rocsolver-path:,custom-target:,address-sanitizer,rm-legacy-include-dir,cmake-arg: --options rhicndgp:v:b: -- "$@")
+  GETOPT_PARSE=$(getopt --name "${0}" --longoptions help,install,codecoverage,clients,no-solver,dependencies,debug,relwithdebinfo,hip-clang,no-hip-clang,compiler:,cmake_install,cuda,use-cuda,cudapath:,installcuda,installcudaversion:,static,relocatable:,rmake_invoked,rocblas:,rocblas-path:,rocsolver-path:,custom-target:,address-sanitizer,cmake-arg: --options rhickndgsb: -- "$@")
```
TorreZuk: Okay so this will also check all rmake.py options it looks like.  The $? -ne 0 was defeated in rocblas install.sh as was removing options from install.sh.   Fine for now

install.sh
```diff
@@ -558,24 +527,16 @@ while true; do
     --cmake-arg)
         cmake_common_options+=("${2}")
         shift 2 ;;
+    --rmake_invoked)
+        rmake_invoked=true
+        shift ;;
     --) shift ; break ;;
     *)  echo "Unexpected command line parameter received; aborting";
         exit 1
         ;;
```
TorreZuk, line 481: Could try to guard only on rmake_invoked but the bi-directional calling support may not be something to keep longer term

TorreZuk (approved):
Okay but are you going to remove all processing of arguments that aren't utilized in the script?

install.sh
daineAMD: Removed any unneeded parsing in install.sh in latest commit.

install.sh
daineAMD: Copied rocblas for now. Would be great to have linux dependency installation somewhere else, install.sh could be very straightforward then. Would be another ticket.

daineAMD:
Looks like all CI has passed ([including Windows](http://mlsejenkinsvm.amd.com/blue/organizations/jenkins/ROCmMathLib%2FhipBLAS/detail/PR-788/4/pipeline)), so should be good to merge.

PR #791, opened 2024-02-20
Develop merge 6.1

TorreZuk:
* brings in 6.1 patches
* reverts the version revert from develop to 6.1 so should be 6.2 compatible


daineAMD (approved):
Looks good, thanks!

PR #797, opened 2024-02-29
her and her2 ILP64

NaveenElumalaiAMD:
resolves: [LWPMLSE-594](https://ontrack-internal.amd.com/browse/LWPMLSE-594)


clients/gtest/blas2/her2_gtest.yaml
```diff
@@ -50,14 +53,45 @@ Tests:
...
+    function:
+      - her2: *single_precision_complex
+
+  - name: her2_64
+    category: pre_checkin
+    arguments:
+      - { N: 2, lda: 2147483649, incx: 1, incy: 1, batch_count: 1 }
+      - { N: 2, lda: 2, incx: 2147483649, incy: 1 , batch_count: 1 }
+      - { N: 2, lda: 2, incx: 1, incy: -2147483649 , batch_count: 1 }
+      - { N: 2, lda: 2, stride_a: 2, incx: 1, batch_count: 666666 }
```
TorreZuk: Talk to Daine and propose a constants in common for this 2147483649 and 666666.  I think we lean toward just large enough to trigger loops rather than int32 overflow unless your tests are fast.   Then we can get just edit common to really test large on the rare occasion but for now leave actual stress tests to rocBLAS by default.   These are problematic enough I think

clients/gtest/blas2/her_gtest.yaml
```diff
@@ -15,6 +15,9 @@ Definitions:
   - &alpha_range
     - [ -0.5, 2.0, 0.0 ]
 
+  - &alpha_range_ILP64
+    - { alpha: -2.5, alphai:  1.5 }
```
TorreZuk: I would tend to smaller like -0.5 and 0.5 as we expect larger sums but these are fine if all passing

clients/gtest/blas2/her_gtest.yaml
```diff
@@ -27,7 +30,7 @@ Tests:
     alpha: *alpha_range
     matrix_size: *size_range
     incx: *incx_range
-    api: [ FORTRAN, C ]
+    api: [ FORTRAN, C, FORTRAN_64, C_64 ]
```
TorreZuk, line 34: We can argue fortran _64 suffices as it invokes C function only would miss double negation bug which would be very low probability.

TorreZuk (approved):
All looks good to me but get Daine's approval before merge, the V2 I am unsure of.

library/src/nvidia_detail/hipblas.cpp
```diff
@@ -10108,6 +10335,135 @@ catch(...)
     return exception_to_hipblas_status();
 }
 
+// her2_64
+hipblasStatus_t hipblasCher2_64(hipblasHandle_t       handle,
+                                hipblasFillMode_t     uplo,
+                                int                   n,
```
daineAMD, line 10341: should be int64_t, along with others further down.

daineAMD (approved):
Looks good aside from comment and testing discussion we had today.

library/include/hipblas.h
```diff
@@ -8639,6 +8639,42 @@ HIPBLAS_EXPORT hipblasStatus_t hipblasZher_v2(hipblasHandle_t         handle,
...
+HIPBLAS_EXPORT hipblasStatus_t hipblasZher_64(hipblasHandle_t             handle,
+                                              hipblasFillMode_t           uplo,
+                                              int64_t                     n,
+                                              const double*               alpha,
+                                              const hipblasDoubleComplex* x,
+                                              int64_t                     incx,
+                                              hipblasDoubleComplex*       AP,
+                                              int64_t                     lda);
+
+HIPBLAS_EXPORT hipblasStatus_t hipblasCher_64_v2(hipblasHandle_t   handle,
```
daineAMD, line 8661: Actually this was changed to be *_v2_64 for other functions so should be changed here too.

TorreZuk:
Windows failure can be ignored but it interesting to see it fail later than first tests with that stack.

PR #799, opened 2024-03-04
ILP64-hbmv-hemv

NaveenElumalaiAMD:
resolves: [LWPMLSE-593](https://ontrack-internal.amd.com/browse/LWPMLSE-593)


clients/include/blas2/testing_hbmv.hpp
```diff
@@ -80,65 +82,75 @@ void testing_hbmv_bad_arg(const Arguments& arg)
...
-            CHECK_HIPBLAS_ERROR(hipblasHbmvFn(
-                handle, uplo, N, K, zero, nullptr, lda, nullptr, incx, one, nullptr, incy));
+            DAPI_CHECK(hipblasHbmvFn,
+                       (handle, uplo, N, K, zero, nullptr, lda, nullptr, incx, one, nullptr, incy));
+        }
+
+        // Check 64-bit API with quick return
+        if(arg.api & c_API_64)
+        {
+            DAPI_CHECK(hipblasHbmvFn, (handle, uplo, N, K, zero, dA, lda, dx, incx, one, dy, incy));
```
daineAMD: Did you mean to test 64-bit sizes here or no?

daineAMD (approved):
I think you should still add some tests to the hbmv_gtest.yaml/hemv_gtest.yaml files to test the 64-bit interface. Can just add C_64 and FORTRAN_64 to some of the existing tests. Also need to add C_64 to the bad_arg tests or else nothing will be actually tested with the 64-bit interface.

clients/include/blas2/testing_hbmv.hpp
NaveenElumalaiAMD: I don't mind adding tests to the `hbmv_gtest.yaml/hemv_gtest.yaml` but thought I could do it after the memory allocation is fixed or else we might see some intermittent failures.

NaveenElumalaiAMD: > Did you mean to test 64-bit sizes here or no?

Not the sizes just the 64-bit API call them and see if returns status success.

NaveenElumalaiAMD:
> I think you should still add some tests to the hbmv_gtest.yaml/hemv_gtest.yaml files to test the 64-bit interface. Can just add C_64 and FORTRAN_64 to some of the existing tests. Also need to add C_64 to the bad_arg tests or else nothing will be actually tested with the 64-bit interface.

I did add the C_64 and FORTRAN_64 to the yaml files but not sure why I am not seeing that commit here.

NaveenElumalaiAMD:
Will go ahead and merge this PR as it has already passed the other CI pipeline already.

PR #811, opened 2024-03-12
tbsv tpsv trsv ILP64

NaveenElumalaiAMD:
resolves: [LWPMLSE-600](https://ontrack-internal.amd.com/browse/LWPMLSE-600)


clients/include/blas2/testing_tpsv.hpp
```diff
@@ -40,8 +40,9 @@ inline void testname_tpsv(const Arguments& arg, std::string& name)
 template <typename T>
 void testing_tpsv_bad_arg(const Arguments& arg)
 {
-    bool FORTRAN       = arg.api == hipblas_client_api::FORTRAN;
-    auto hipblasTpsvFn = FORTRAN ? hipblasTpsv<T, true> : hipblasTpsv<T, false>;
+    bool FORTRAN          = arg.api == hipblas_client_api::FORTRAN;
+    auto hipblasTpsvFn    = FORTRAN ? hipblasTpsv<T, true> : hipblasTpsv<T, false>;
+    auto hipblasTpsvFn_64 = FORTRAN_64 ? hipblasTpsv_64<T, true> : hipblasTpsv_64<T, false>;
```
TorreZuk: Well conditional with constant should give warning as I don't see bool FORTRAN_64.   Should check the enum if we did use FORTRAN AND FORTRAN_64, probably best to replace the bool FORTRAN as well.  Both could use the fortran bit mask instead so they are both the same arg.api & c_FORTRAN ?   if same constants used as rocblas.

clients/include/blas2/testing_tpsv.hpp
```diff
@@ -58,52 +59,61 @@ void testing_tpsv_bad_arg(const Arguments& arg)
...
 }
 
 template <typename T>
 void testing_tpsv(const Arguments& arg)
 {
-    bool FORTRAN       = arg.api == hipblas_client_api::FORTRAN;
-    auto hipblasTpsvFn = FORTRAN ? hipblasTpsv<T, true> : hipblasTpsv<T, false>;
+    bool FORTRAN          = arg.api == hipblas_client_api::FORTRAN;
+    auto hipblasTpsvFn    = FORTRAN ? hipblasTpsv<T, true> : hipblasTpsv<T, false>;
+    auto hipblasTpsvFn_64 = FORTRAN_64 ? hipblasTpsv_64<T, true> : hipblasTpsv_64<T, false>;
```
TorreZuk: Same advice as earlier

TorreZuk (approved):
Just spotted the function pointer conditional for  _64 FORTRAN_64 that you can discuss with @daineAMD

clients/include/blas2/testing_tpsv.hpp
NaveenElumalaiAMD: Thanks for pointing it out, I will do the fix now. Sure, @daineAMD could advise about this change.

daineAMD: Enum is:
```
typedef enum hipblas_client_api_
{
    C          = 0,
    C_64       = 1,
    FORTRAN    = 2,
    FORTRAN_64 = 3
} hipblas_client_api;
```
I'm fine with `arg.api & c_API_FORTRAN` for both, can change them all afterward.

library/include/hipblas.h
```diff
@@ -12589,6 +12589,29 @@ HIPBLAS_EXPORT hipblasStatus_t hipblasDspr2_64(hipblasHandle_t   handle,
                                                const double*     y,
                                                int64_t           incy,
                                                double*           AP);
+
+// 64-bit interface
+HIPBLAS_EXPORT hipblasStatus_t hipblasSspr2Batched_64(hipblasHandle_t    handle,
```
daineAMD: I think this (and the strieded_batched versions) were moved around accidentally, this should be around L12709, for example. I think this is what's causing the CI documentation failures.

NaveenElumalaiAMD: Thanks for catching, something went wrong with the rebase.

PR #821, opened 2024-03-20
Fix hipBLAS host memory allocation

NaveenElumalaiAMD:
resolves: [LWPMLSE-607](https://ontrack-internal.amd.com/browse/LWPMLSE-607)

Summary of proposed changes:
- If the host memory allocation exceeds the available RAM then we skip the test similar to rocBLAS.


clients/include/hipblas_init.hpp
```diff
@@ -69,6 +69,285 @@ typedef enum hipblas_matrix_type_
...
 } hipblas_matrix_type;
 
+/* ============================================================================================ */
+
+/* ============================================================================================ */
+/*! \brief  matrix/vector initialization: */
+// for vector x (M=1, N=lengthX, lda=incx);
+// for complex number, the real/imag part would be initialized with the same value
+template <typename T>
+void hipblas_init(host_vector<T>& A,
```
daineAMD, line 79: Are there any changes here other than importing the functions from utility.h?

clients/gtest/hipblas_gtest_main.cpp
```diff
@@ -148,27 +148,39 @@ class ConfigurableEventListener : public TestEventListener
...
+        if(result.type() == TestPartResult::kSkip)
+        {
+            ++skipped_tests;
+
+            if(strstr(result.message(), LIMITED_RAM_STRING))
+            {
+                if(showInlineSkips)
+                    std::cout << "Skipped test due to limited RAM environment." << std::endl;
+            }
+            else if(strstr(result.message(), LIMITED_VRAM_STRING))
```
daineAMD, line 160: We'll need to wrap our testing device allocations with `CHECK_DEVICE_ALLOCATION` to get this feature to work, is that right? I see we already have that defined but it isn't used anywhere.

clients/common/host_alloc.cpp
```diff
@@ -0,0 +1,229 @@
...
+    if(host_mem_safe(size))
+    {
+        void* ptr = malloc(size);
+
+        static int value = -1;
+
+        static auto once = false;
+        if(!once)
+        {
+            auto* alloc_byte_str = getenv("HIPBLAS_CLIENT_ALLOC_FILL_HEX_BYTE");
```
daineAMD, line 194: This is just used for our own debugging, I suppose? @TorreZuk this question might be aimed at you.

daineAMD (commented):
Think it looks good generally, but should definitely get another review from Torre since he made original changes in rocBLAS I believe.
Needs addition of `CHECK_DEVICE_ALLOCATION`, maybe just start with a single function for this PR and we can add it throughout afterwards. Assuming you've tested out both `HIPBLAS_CLIENT_RAM_GB_LIMIT` and `HIPBLAS_CLIENT_ALLOC_AVAILABLE` with large tests. Can add back some of l1 stress int64 tests here to test it out on CI too if you want.

clients/common/host_alloc.cpp
```diff
@@ -0,0 +1,229 @@
...
+    {
+        return true;
+    }
+
+    constexpr size_t threshold = 100 * 1024 * 1024; // 100 MB
+
+    static size_t client_ram_limit = 0;
+
+    static int once = [&] {
+        auto* alloc_limit = getenv("HIPBLAS_CLIENT_RAM_GB_LIMIT");
```
TorreZuk, line 144: Rather than this library specific perhaps better to ask CI to set ROCM_CI_CLIENT_RAM_GB_LIMIT as they just added ROCM_CI_RAM_GB_LIMIT for building.  Get away from a variable per math library....

clients/include/hipblas_init.hpp
TorreZuk: Yes not sure how this is related to this PR?   A different PR maybe....

TorreZuk (approved):
Asked some questions to address, but if your testing you can get it SKIPPING some artificially high RAM tests, and you pulled everything applicable from rocBLAS then fine by me.  The env variable limit you can test locally but we can get separate CI PR for the new env variable (and make rocblas change too).

clients/common/host_alloc.cpp
NaveenElumalaiAMD: I checked it locally and it seems to be working fine. Sure, when we remove the ROCBLAS_CLIENT_RAM_GB_LIMIT will remove the HIPBLAS_CLIENT_RAM_GB_LIMIT as well, or do you want me to remove this feature from this PR?

TorreZuk: Fine to leave in for now and change later with rocblas, does the groovy set this var?

NaveenElumalaiAMD: No the groovy doesn't set it.

TorreZuk: Well set the env in the groovy so your code is tested by CI.  Can be this PR or another as simple additions

PR #826, opened 2024-03-22
fix some bad merges

daineAMD:
Some bad merges happened after changing a lot of internal function names, this should fix that.


TorreZuk (approved):
LGTM if tests pasing.

PR #829, opened 2024-03-25
fix install.sh --help option and remove warnings for invalid params

daineAMD:
- remove unwanted expansion of some info in ./install.sh --help 
- remove warnings invalid params to install.sh as some will be passed straight through to rmake.py


TorreZuk (approved):
After display_help call in rocBLAS we also
python3 ./rmake.py --help
exit 0

daineAMD:
> After display_help call in rocBLAS we also python3 ./rmake.py --help exit 0

Made a couple further changes, including adding this suggestion.

PR #830, opened 2024-03-26
Memory allocators for vector and matrix variants

NaveenElumalaiAMD:
Vector and matrix memory allocators similar to rocBLAS


TorreZuk (approved):
If all tests passing LGTM

daineAMD (approved):
May want to use some of the new types in some of the testing_*.hpp files before merge

NaveenElumalaiAMD:
> May want to use some of the new types in some of the testing_*.hpp files before merge

I have these changes in my next PR (locally built without any issues) once this goes in I will open that one.

PR #833, opened 2024-03-27
back out while fixing

TorreZuk:
* actually have AOCL 4.2 installed now so back out and fix


NaveenElumalaiAMD (approved):
In the pre-checkin pipeline i see it is linking to aocc-4.0 rather than aocc-4.1.

TorreZuk:
> In the pre-checkin pipeline i see it is linking to aocc-4.0 rather than aocc-4.1.

Sure whatever fallback is installed is fine.  The 4.1 is the next candidate

PR #834, opened 2024-03-28
use flame ILP64 allow mult. def.

TorreZuk:
* retest AOCL 4.2 with different usage for solver


NaveenElumalaiAMD (approved):
LGTM! It is good to see solver tests passing with aocc-4.2.0.

PR #835, opened 2024-04-01
Reducing the number and sizes of tests

daineAMD:
Greatly reduce the number of tests and the size of the tests in hipBLAS. We don't need to test a bunch of sizes here, just one or two to make sure the API is passing in the arguments correctly. I don't think the changes here should really reduce coverage w.r.t. hipblas.

On my local gfx906 (without solver tests):

| executable | develop # tests | develop time | new # tests | new time |
| --- | --- | --- | --- | --- |
| hipblas-test | 702893 | 84m | 39769 | 115s |
| hipblas_v2-test | 351456 | 38m | 19886 | 60s |

### DIFF
--- a/clients/gtest/blas1/asum_gtest.yaml
+++ b/clients/gtest/blas1/asum_gtest.yaml
@@ -3,13 +3,13 @@ include: hipblas_common.yaml
 
 Definitions:
   - &N_range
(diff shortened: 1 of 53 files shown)


TorreZuk (approved):
It makes sense to me, a single valid size that does work and all values are different to catch any mapping error seems fine

PR #836, opened 2024-04-02
Testing rocsolver, rocsparse, rocprim dependencies with noSolver label

daineAMD:


daineAMD:
Flag seems to work as intended - doesn't look like CI queries jobs for rocSOLVER or its dependencies.

PR #839, opened 2024-04-04
Add LLVM_DEFAULT_TARGET_TRIPLE to find new location of -lomp

estewart08:
Summary of proposed changes:
An upstream llvm change enables ```LLVM_ENABLE_PER_TARGET_RUNTIME_DIR``` by default for the openmp
build. This installs the openmp libraries into ```/opt/rocm-ver/llvm/lib/x86_64-unknown-linux-gnu``` instead of
```/opt/rocm-ver/llvm/lib```. Currenty, hipBLAS only looks in /lib. Prepend lib/x86_64-unknown-linux-gnu to ```-L``` and ```--rpath``` to avoid a linker error when the upstream change lands in ```amd-staging```.

### DIFF
--- a/clients/CMakeLists.txt
+++ b/clients/CMakeLists.txt
@@ -118,11 +118,23 @@ if( BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
   # if there is no omp.h to find the client compilation will fail and this should be obvious, used to be REQUIRED
   find_package(OpenMP)
 
+  execute_process (
+    COMMAND bash -c "${HIP_CLANG_ROOT}/bin/clang --version | grep -oP '(?<=Target: ).+' | tr -d '\n'"
+    OUTPUT_VARIABLE LLVM_DEFAULT_TARGET_TRIPLE)
+
   if (TARGET OpenMP::OpenMP_CXX)
     set( COMMON_LINK_LIBS "OpenMP::OpenMP_CXX")
     if(HIP_PLATFORM STREQUAL amd)
+      if (NOT WIN32)
+        if(LLVM_DEFAULT_TARGET_TRIPLE)
+          list( APPEND COMMON_LINK_LIBS "-L\"${HIP_CLANG_ROOT}/lib/${LLVM_DEFAULT_TARGET_TRIPLE}\"")
+        endif()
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: clients/CMakeLists.txt [hunk 1]
```diff
@@ -121,8 +121,19 @@ if( BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
   if (TARGET OpenMP::OpenMP_CXX)
     set( COMMON_LINK_LIBS "OpenMP::OpenMP_CXX")
     if(HIP_PLATFORM STREQUAL amd)
+      if (NOT WIN32)
+        execute_process (
+          COMMAND bash -c "${HIP_CLANG_ROOT}/bin/clang --version | grep -oP '(?<=Target: ).+' | tr -d '\n'"
+          OUTPUT_VARIABLE LLVM_DEFAULT_TARGET_TRIPLE)
+        if(LLVM_DEFAULT_TARGET_TRIPLE)
+          list( APPEND COMMON_LINK_LIBS "-L\"${HIP_CLANG_ROOT}/lib/${LLVM_DEFAULT_TARGET_TRIPLE}\"")
+        endif()
```
TorreZuk, line 132: I assume you have no way to add this into OpenMP::OpenMP_CXX  as that would be cleaner approach.   I thought we were moving to use /bin/amdclang as well.   Otherwise appears fine

estewart08: The ```find_package(OpenMP)``` is actually using FindOpenMP.cmake from the cmake install itself. From what I have seen, it is finding libgomp.so from a system package and not ROCm. Passing a custom ```CMAKE_PREFIX_PATH``` will not work either for this as they turn off these search paths in FindOpenMP.cmake. This is probably why the ```-L /lib``` had to be added in the first place. The openmp install does not have a config file to query either.

As far as using ```amdclang```, I can make that change.

TorreZuk (approved):
Okay, it is getting messy but acceptable I guess

daineAMD (approved):
CI looks happy with it, so fine by me.

estewart08:
Closed in favor of: https://github.com/ROCm/hipBLAS/pull/1038

PR #841, opened 2024-04-08
Adding flags to choose compiler in hipblas PRs

daineAMD:


TorreZuk (approved):
LGTM if works

daineAMD:
The flag seems to work fine. There's a problem with the hipcc build, but I'll resolve that in another PR.

PR #842, opened 2024-04-09
Fix hipcc build for C example

daineAMD:
Not sure why this was originally set to CXX, so maybe I'm missing something.
Develop passes on my ROCm 6.0 machine as-is, but fails on the docker image currently in use by CI when built with hipcc. With this change, it seems to work on both.

### DIFF
--- a/clients/samples/CMakeLists.txt
+++ b/clients/samples/CMakeLists.txt
@@ -47,7 +47,7 @@ if(HIP_PLATFORM STREQUAL amd)
 
   add_executable( hipblas-example-c example_c.c ${hipblas_samples_common} )
   # We test for C99 compatibility in the example_c.c test
-  set_source_files_properties(example_c.c PROPERTIES LANGUAGE CXX)
+  set_source_files_properties(example_c.c PROPERTIES LANGUAGE C)
   set_source_files_properties(example_c.c PROPERTIES COMPILE_FLAGS "-xc -std=c99")
 endif()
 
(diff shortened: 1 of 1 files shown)


clients/samples/CMakeLists.txt
```diff
@@ -47,7 +47,7 @@ if(HIP_PLATFORM STREQUAL amd)
 
   add_executable( hipblas-example-c example_c.c ${hipblas_samples_common} )
   # We test for C99 compatibility in the example_c.c test
-  set_source_files_properties(example_c.c PROPERTIES LANGUAGE CXX)
+  set_source_files_properties(example_c.c PROPERTIES LANGUAGE C)
   set_source_files_properties(example_c.c PROPERTIES COMPILE_FLAGS "-xc -std=c99")
```
TorreZuk: -xc told the compiler it was C language.     Are you setting CC compile and tool chain?  This may have been simpler to get the cxx compiler just to compile as C99.  Some C compiler may not like -xc but the -std-c99 probably okay.

TorreZuk (approved):
If hipcc and gcc happy

TorreZuk:
Looks like a lot of issues in last run, picking out the C example, if you go with C compiler then all cmake C flags should match CXX flags, maybe that is why we went with using CXX compiler and just told it file was C language:
/usr/bin/ld: CMakeFiles/hipblas-example-c.dir/example_c.c.o: relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a PIE object; recompile with -fPIE

daineAMD:
Will test out on cuda and make further changes if all passes, just waiting on docker for now.

daineAMD:
Passed CI with both gcc and hipcc, along with cuda. Tested manually as well. The multicompiler job is broken but for something unrelated.

PR #847, opened 2024-04-12
Revert "Properly locate libomp.so when installed in clang target subdirectory."

amcamd:
Reverts ROCm/hipBLAS#840

revert to avoid build break in [SWDEV-456451](https://ontrack-internal.amd.com/browse/SWDEV-456451)

### DIFF
--- a/clients/CMakeLists.txt
+++ b/clients/CMakeLists.txt
@@ -73,7 +73,7 @@ if( NOT DEFINED CMAKE_CONFIGURATION_TYPES AND NOT DEFINED CMAKE_BUILD_TYPE )
 endif()
 
 # This project may compile dependencies for clients
-project( hipblas-clients LANGUAGES C CXX Fortran )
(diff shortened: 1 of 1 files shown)


TorreZuk (approved):
I think fine to backout while a proper fix is analyzed

estewart08:
Seems like they don't have consistent packages on the rpm based distros (missing  `rocm-llvm-devel`). Hardcoding a a guess at -L is not great either, but it will get you unblocked. Unfortunately, when libomp is not at that location, moving in the near future, it will fail. The find_package(OpenMP) is not doing much here either when compiling with gcc as it cannot use libgomp.

estewart08 (approved):
I will have to come up with a slightly different solution when rocm-llvm-dev/devel is not installed.

amcamd:
Thanks @estewart08 , Note that we are not reverting the changes in rocBLAS.

PR #848, opened 2024-04-14
Remove immintrin header

cgmb:
This header provides x86 intrinsics, but is not used within the hipblas clients.


cgmb:
My mistake. It's only unused if the hipBLAS clients are built using `__half` as the float type. Otherwise, it is used for `_cvtss_sh`.

PR #849, opened 2024-04-14
Link hip::host as PUBLIC

cgmb:
The hipblas library includes HIP headers in its public headers, so it should link hip::host as PUBLIC. This will ensure that the HIP runtime include paths are passed to any library that links against roc::hipblas.


daineAMD (approved):

cgmb:
The link failures on CI in this PR appear to be the same as on develop.

PR #852, opened 2024-04-18
Fix hipblasMath_t missing when using hipSOLVER

cgmb:
The hipblasMath_t type declaration was inside the side mode guard, preventing the math enums from being declared if hipSOLVER declared hipblasSideMode_t. This should probably be cherry-picked into 6.1.1.


TorreZuk (approved):
Fix looks good.  
I vote that this gets replaced with a common header package in the future

tfalders (approved):

TorreZuk:
@daineAMD can you merge this if you accept it, as best to cherry pick it to 6.2 release-staging while it requires no extra testing.

PR #853, opened 2024-04-19
Revamp how libomp.so is found

estewart08:
An upstream llvm change enables
LLVM_ENABLE_PER_TARGET_RUNTIME_DIR by default for the openmp build. This installs the openmp libraries into
`/opt/rocm-ver/llvm/lib/x86_64-unknown-linux-gnu` instead of `/opt/rocm-ver/llvm/lib`. Currenty, hipBLAS only uses /lib.

Since hipBLAS uses gcc by default, `find_package(OpenMP)` will locate libgomp from the gnu installation. We would prefer to use libomp from the ROCm install. Query amdclang to find LLVM_TARGET_TRIPLE and use as a suffix for `find_library(omp)`.

On the other hand, the user can choose to use hipcc, which allows `find_package(OpenMP)` to properly locate libomp inside ROCm llvm.

To include legacy support, I left the hardcoded `-L HIP_CLANG_ROOT` as a fallback.

This would be much simpler if hipBLAS defaulted to hipcc as the cmake compiler.

Summary of proposed changes:
- If compiler is gcc, query amdclang for LLVM_TARGET_TRIPLE and use as suffix for `find_library(omp)`.
- If compiler is hipcc, use `find_package(OpenMP)`.
- If cmake cannot find libomp.so then fallback to using `-L HIP_CLANG_ROOT/lib`.
- Allow proper detection of new lib installation subdirectory (i.e. x86_64-unknown-linux-gnu).
- Removed dependency on `find_package(LLVM)` due to `rocm-llvm-dev/devel` not being installed by default on various operating systems, which would not have the cmake config files.

### DIFF
--- a/clients/CMakeLists.txt
+++ b/clients/CMakeLists.txt
@@ -136,14 +136,15 @@ if( BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
     endif()
   endif()
 
-  if(HIP_PLATFORM STREQUAL amd)
-    if(LIBOMP_FOUND)
-      set( COMMON_LINK_LIBS ${LIBOMP_FOUND})
-      if (NOT WIN32)
-        list( APPEND COMMON_LINK_LIBS "-Wl,-rpath=${LIBOMP_PATH}")
-      endif()
-    # If cmake cannot find the proper libomp.so, use HIP_CLANG_ROOT as fallback.
-    else()
+  if(LIBOMP_FOUND)
+    # LIBOMP_FOUND will contain full path to libomp.so
+    set( COMMON_LINK_LIBS ${LIBOMP_FOUND})
+    if(HIP_PLATFORM STREQUAL amd AND NOT WIN32)
+      list( APPEND COMMON_LINK_LIBS "-Wl,-rpath=${LIBOMP_PATH}")
+    endif()
+  # If cmake cannot find the proper libomp.so, use HIP_CLANG_ROOT as fallback.
+  else()
+    if(HIP_PLATFORM STREQUAL amd)
       # Add library search paths.
       list( APPEND COMMON_LINK_LIBS "-L\"${HIP_CLANG_ROOT}/lib/${LLVM_TARGET_TRIPLE}\"")
       list( APPEND COMMON_LINK_LIBS "-L\"${HIP_CLANG_ROOT}/lib\"")
--- a/clients/CMakeLists.txt
+++ b/clients/CMakeLists.txt
@@ -124,13 +124,17 @@ if( BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
     endif()
   # Support for using gnu as cmake compiler.
   else()
+    # Discover path to amdclang
+    get_filename_component(AMDCLANG_PATH "${HIP_HIPCC_EXECUTABLE}/../amdclang" REALPATH)
+    get_filename_component(AMDCLANG_DIR "${AMDCLANG_PATH}" DIRECTORY)
+    set(AMDCLANG_LIBDIR "${AMDCLANG_DIR}/../lib")
     # libomp.so may be in a target subdirectory directory.
     execute_process(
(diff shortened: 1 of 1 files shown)


clients/CMakeLists.txt
```diff
@@ -116,13 +116,40 @@ if( BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
...
+      OUTPUT_VARIABLE LLVM_TARGET_TRIPLE
+    )
+    find_library(LIBOMP_FOUND omp PATHS ${HIP_CLANG_ROOT}/lib PATH_SUFFIXES ${LLVM_TARGET_TRIPLE} NO_DEFAULT_PATH)
+    if (LIBOMP_FOUND)
+      get_filename_component(LIBOMP_PATH "${LIBOMP_FOUND}" PATH)
+    endif()
+  endif()
 
-  if (TARGET OpenMP::OpenMP_CXX)
-    set( COMMON_LINK_LIBS "OpenMP::OpenMP_CXX")
```
TorreZuk, line 122: So aren't we missing the OpenMP::OpenMP_CXX link request on CUDA backend?

estewart08: True, I can move `set( COMMON_LINK_LIBS ${LIBOMP_FOUND})` outside of the amd guard.

clients/CMakeLists.txt
```diff
@@ -116,13 +116,41 @@ if( BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
 
   # if it fails to find OpenMP compile and link flags in strange configurations it can just use non-parallel reference computation
   # if there is no omp.h to find the client compilation will fail and this should be obvious, used to be REQUIRED
-  find_package(OpenMP)
+  if( CMAKE_CXX_COMPILER MATCHES ".*/hipcc$" )
+    find_package(OpenMP)
```
TorreZuk, line 120: Similar question we shouldn't require NON hipcc to find_package(OpenMP) so can this just be moved two lines above?   Sorry I meant other than hipcc we still want find_package to set cmake target etc.

estewart08: When the cmake compiler is hipcc `find_package` finds the correct ROCm libomp.so. Otherwise, gcc will get libgomp.so from gcc installation. I would rather get the `libomp` from ROCm to be consistent. With your current configuration cmake adds `libgomp.so` with `-L/rpaths -lomp` to the ROCm library. So I am fairly certain libgomp is ignored.

TorreZuk:
Hi @estewart08 are you still active on this?

estewart08:
I am going to pick this effort back up, so I would prefer it to stay open for now.

TorreZuk:
@estewart08 if picking this up again for next release time is ticking....

TorreZuk:
@estewart08 you mentioned you were going to continue this work.  Any ETA?

estewart08:
I am currently looking into providing an openmp-config.cmake file that will provide the proper location of the library. No definitive ETA at the moment.

TorreZuk:
> I am currently looking into providing an openmp-config.cmake file that will provide the proper location of the library. No definitive ETA at the moment.

@estewart08 @stellaraccident  This would be a big help as we are now seeing issues with tighter redhat 10 rpath restrictions so avoiding the workaround by getting full cmake config could solve newer problems.

stellaraccident:
> > I am currently looking into providing an openmp-config.cmake file that will provide the proper location of the library. No definitive ETA at the moment.
> 
> @estewart08 @stellaraccident This would be a big help as we are now seeing issues with tighter redhat 10 rpath restrictions so avoiding the workaround by getting full cmake config could solve newer problems.

I don't have full state on the set of constraints that need to be solved, but I've definitely developed... Let's call it experience... On this issue and agree we need to improve the situation. Agreed that some form of standalone cmake config for this seems like the right answer. If we want to look at details, maybe look at this in upcoming office hours.

estewart08:
Closed in favor of: https://github.com/ROCm/hipBLAS/pull/1038

PR #858, opened 2024-05-01
Fix hipblasMath_t missing when using hipSOLVER (#852)

daineAMD:
The hipblasMath_t type declaration was inside the side mode guard, preventing the math enums from being declared if hipSOLVER declared hipblasSideMode_t.


TorreZuk (approved):
LGTM

PR #872, opened 2024-05-30
l3 64-bit interface enable

daineAMD:
Enables L3 64-bit interfaces with AMD backend, along with gemm_ex. Tested manually against Naveen's PR, will remove noCI label once that's in rocblas develop.


clients/include/blas_ex/testing_gemm_batched_ex.hpp
```diff
@@ -124,184 +128,167 @@ void testing_gemm_batched_ex_bad_arg(const Arguments& arg)
...
+                         one,
+                         nullptr,
+                         cType,
+                         c_i32_overflow,
+                         c_i32_overflow,
+                         computeType,
+                         algo, flags));
+                        DAPI_EXPECT((arg.api & c_API_64) ? HIPBLAS_STATUS_SUCCESS
+                                             : HIPBLAS_STATUS_INVALID_VALUE,
+                        hipblasGemmBatchedExFn,
```
TorreZuk: Space or is clang format off?

TorreZuk (approved):
Very good.   In the detail files once V2 goes away we may want to locally provide a bunch of conversion inline constructors to do all the casting as that seems a lot of the code.

daineAMD:
Only CI failures appear to be rotm failures

PR #875, opened 2024-06-04
Fix rotm_strided_batched CI failures

NaveenElumalaiAMD:


clients/include/blas1/testing_rotm_strided_batched.hpp
```diff
@@ -176,7 +176,15 @@ void testing_rotm_strided_batched(const Arguments& arg)
     hipblas_init_vector(hdata, arg, hipblas_client_alpha_sets_nan, false);
 
     for(int64_t b = 0; b < batch_count; b++)
+    {
+        T* hparam_ptr = hparam[b];
+
+        // generating simply one set of hparam which will not be appropriate for testing
+        // that it zeros out the second element of the rotm vector parameter
+        memset(hparam_ptr, 0, 5 * sizeof(T));
```
TorreZuk: Ahhh okay we aren't zero initialized.  It is a single memory block so you should pull the memset outside the block and do the full * batch_count

TorreZuk (approved):

clients/include/blas1/testing_rotm_strided_batched.hpp
TorreZuk: Or you could add a zero( ) method to the helper and do the memset inside if we expect it elsewhere.

TorreZuk: Sorry dug deeper as was curious, think you just want to fix host_strided_batched_vector internal allocation it is an outlier, it should use host_calloc_throw instead of host_malloc_throw (see others).  Check rocblas too please.

NaveenElumalaiAMD: In rocBLAS, the initial design I did was to use the `host_calloc_throw` but for some reasons we went with `host_malloc_throw` (may be it was faster) not exactly sure why. I can change it back to `host_calloc_throw` if everybody is okay.

TorreZuk: Yes malloc will be slightly faster, but may as well be consistent.    If we change all to malloc then need zero( ) methods for cases where not all data is initialized.

PR #873, opened 2024-05-31
merge release-staging/rocm-rel-6.2 into develop and hipconfig path change

daineAMD:
merge release-staging/rocm-rel-6.2 into develop plus changing path install.sh looks for hipconfig to potentially fix static CI failures; not sure if it'll work. Could potentially change CI to just use rmake.py instead.

PR #877, opened 2024-06-06
cmake: don't set CMAKE_C_COMPILER, as hipBLAS is a CXX project

evetsso:
See https://github.com/ROCm/rocPRIM/pull/567 for a description of the errors solved by this change.

### DIFF
--- a/toolchain-linux.cmake
+++ b/toolchain-linux.cmake
@@ -14,20 +14,14 @@ if (NOT DEFINED ENV{CXX} AND NOT CMAKE_CXX_COMPILER)
   set(CMAKE_CXX_COMPILER "${rocm_bin}/hipcc")
 endif()
 
-if (NOT DEFINED ENV{CC} AND NOT CMAKE_C_COMPILER)
-  set(CMAKE_C_COMPILER "${rocm_bin}/hipcc")
-endif()
-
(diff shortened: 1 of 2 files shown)


daineAMD (approved):
LGTM as long as CI passes, looks like cuda CI failed due to docker issues.
I can merge back into develop afterward.

TorreZuk (approved):
LGTM

daineAMD:
I think we'll need the change from https://github.com/ROCm/hipBLAS/pull/873/commits/5f23b35283be7110d4a82d7253f82bc0a0202851 to go into release-staging/rocm-rel-6.2 as well to get static builds to pass, so I expect static CI to fail here.

I'm ok if you go ahead and merge this one even with the static failure and I can create a PR afterwards to include that change. Otherwise you can make that change here too if you want.

TorreZuk:
Will re-run with noSOLVER to see if at least something can build.

evetsso:
> Will re-run with noSOLVER to see if at least something can build.

Thanks.  FWIW, I've been able to manually build with my own rocSOLVER.

TorreZuk:
Re-run in CI passed on one OS without solver (non static) so IMO you are safe to merge.

evetsso:
Oh.. seems I don't have permission to merge.  @TorreZuk , @daineAMD I think one of you can probably do it?

PR #878, opened 2024-06-06
Fixes for static build

daineAMD:
With this and #877, static builds should work. Will delete #873 and have simple merge back from 6.2 -> develop after these are in.


TorreZuk (approved):
I saw stale information for my last comment, it actually looks fine to merge.

PR #881, opened 2024-06-12
Change default compiler to amdclang++

daineAMD:
hipBLAS doesn't have compiler heuristics like rocBLAS, so straightforward changes. Will do a little bit of manual performance testing but don't expect any perf changes in hipBLAS.

### DIFF
--- a/clients/gtest/CMakeLists.txt
+++ b/clients/gtest/CMakeLists.txt
@@ -195,16 +195,6 @@ if(HIP_PLATFORM STREQUAL amd)
     target_link_libraries( hipblas_v2-test PRIVATE hip::${CUSTOM_TARGET} )
   endif( )
 
-  if( CMAKE_CXX_COMPILER MATCHES ".*/hipcc$" )
-# hip-clang needs specific flag to turn on pthread and m
-    target_link_libraries( hipblas-test PRIVATE -lpthread -lm )
-    target_link_libraries( hipblas_v2-test PRIVATE -lpthread -lm )
-
-    if(BUILD_ADDRESS_SANITIZER)
-      target_link_libraries( hipblas-test PRIVATE -fuse-ld=lld -lgfortran )
-      target_link_libraries( hipblas_v2-test PRIVATE -fuse-ld=lld -lgfortran )
-    endif()
-  endif()
 else( )
   target_include_directories( hipblas-test
     PRIVATE
--- a/clients/samples/CMakeLists.txt
+++ b/clients/samples/CMakeLists.txt
@@ -49,6 +49,8 @@ if(HIP_PLATFORM STREQUAL amd)
 # We test for C99 compatibility in the example_c.c test
 set_source_files_properties(example_c.c PROPERTIES LANGUAGE CXX)
 set_source_files_properties(example_c.c PROPERTIES COMPILE_FLAGS "-xc -std=c99")
+# needed for hipcc to find hip includes for C example
+target_compile_options(hipblas-example-c PRIVATE -I${HIP_INCLUDE_DIRS})
 
 # We test for C++11 compatibility in one of the samples
 #set_source_files_properties(example_sgemm_strided_batched.cpp PROPERTIES COMPILE_FLAGS "-std=c++11")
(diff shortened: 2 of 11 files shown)


rmake.py
```diff
@@ -61,7 +61,7 @@ def parse_args():
                         help='Code coverage build. Requires Debug (-g|--debug) or RelWithDebInfo mode (-k|--relwithdebinfo), (optional, default: False)')
 
     parser.add_argument(      '--compiler', type=str, required=False, default=None, dest='compiler',
-                        help='Spcify path to host compiler.')
+                        help='Spcify path to host compiler. Default is amdclang++.')
```
TorreZuk: Might as well fix spelling

rmake.py
```diff
@@ -118,7 +118,7 @@ def parse_args():
                         help='[DEPRECATED] Linux only: Build hipBLAS using hipcc compiler. Deprecated, use --compiler instead.')
 
     parser.add_argument(      '--no-hip-clang', dest='use_hipcc_compiler', required=False, default=True, action='store_false',
-                        help='[DEPRECATED] Linux only: Build hipBLAS with g++ compiler instead of hipcc compiler. Deprecated, use --compiler instead.')
+                        help='[DEPRECATED] Linux only: Build hipBLAS with default amdclang++ compiler instead of hipcc compiler. Deprecated, use --compiler instead.')
```
TorreZuk: I wouldn't change a deprecated flag, it was to use g++ so still valid  as is (hipcc and amdclang are the same compiler)

rmake.py
```diff
@@ -268,13 +268,10 @@ def config_cmd():
     if args.compiler is not None:
         cmake_options.append(f"-DCMAKE_CXX_COMPILER={args.compiler}")
 
-    # windows defaults to clang++ in toolchain
+    # amdclang++ default in linux toolchain, clang++ in windows toolchain
     if os.name != "nt" and args.compiler is None:
         if args.use_hipcc_compiler:
             cmake_options.append(f"-DCMAKE_CXX_COMPILER={rocm_path}/bin/hipcc")
```
TorreZuk: This is the line you need to change.  Probably should change use_hipcc_compiler to use_amdclang_compiler?   If they want to actually use hipcc they can use --compiler flag

TorreZuk (approved):
Changes requested but minor

clients/samples/CMakeLists.txt
```diff
@@ -63,7 +61,7 @@ if( NOT TARGET hipblas )
 endif( )
 
 list (APPEND hipblas-example-executables hipblas-example-sscal hipblas-example-scal-ex-v2 hipblas-example-strmm hipblas-example-sgemm hipblas-example-sgemm-strided-batched hipblas-example-gemmEx_v2 hipblas-example-hip-complex-her2 hipblas-example-hgemm-half hipblas-example-c ${sample_list_fortran} )
-if( CMAKE_CXX_COMPILER MATCHES ".*/hipcc$" )
+if( CMAKE_CXX_COMPILER MATCHES ".*/hipcc$" OR CMAKE_CXX_COMPILER MATCHES ".*/amdclang\\+\\+$")
```
TorreZuk, line 66: Are you using hipcc on windows?  Otherwise you can try:
CMAKE_CXX_COMPILER_ID MATCHES "Clang" for more general conditional

clients/gtest/CMakeLists.txt
```diff
@@ -195,16 +195,6 @@ if(HIP_PLATFORM STREQUAL amd)
...
-  if( CMAKE_CXX_COMPILER MATCHES ".*/hipcc$" )
-# hip-clang needs specific flag to turn on pthread and m
-    target_link_libraries( hipblas-test PRIVATE -lpthread -lm )
-    target_link_libraries( hipblas_v2-test PRIVATE -lpthread -lm )
-
-    if(BUILD_ADDRESS_SANITIZER)
-      target_link_libraries( hipblas-test PRIVATE -fuse-ld=lld -lgfortran )
-      target_link_libraries( hipblas_v2-test PRIVATE -fuse-ld=lld -lgfortran )
-    endif()
-  endif()
```
TorreZuk, line 207: Is this duplicated and already set in another file?    Would think they may still be required for amdclang ....

daineAMD: I was able to build locally with `./install.sh -cn --compiler=hipcc` with these changes. Testing out address sanitizer on different arch now and will see how CI goes

daineAMD: Address sanitizer was broken before this commit which may fix it, seems to be broken in rocblas as well. I'll see if I can fix it in rocblas tomorrow and try again in hipblas, but guessing it isn't used by anyone since the option is broken in our build scripts.

PR #889, opened 2024-07-08
Start using hipblas-common package

bnemanich:
resolves: hipBLASLt build dependency on hipBLAS by utilizing the hipBLAS-common package.

Summary of proposed changes:
- Removes datatypes that exist within hipBLAS-common package from hipBLAS
- Adds hipblas-common as a dependency to hipBLAS.


daineAMD (changes requested):
- Should add a note in install.rst referencing the new dependency. Also potentially in functions.rst to reference the external types
- Need to get [Windows CI](https://mlsejenkinsvm.amd.com/blue/organizations/jenkins/ROCmMathLib%2FhipBLAS/detail/PR-889/2/pipeline/33) dependency available so windows CI isn't broken in hipBLAS. It looks to me like the packaging already exists in the hipblas-common repo, just needs CI running there, and the dependency added for hipBLAS CI
- Rerunning [pre-checkin CI](http://math-ci.amd.com/blue/organizations/jenkins/main%2Fprecheckin%2FhipBLAS/detail/PR-889/3/pipeline/442) to see if it passes
Edit: also [cuda CI is failing](http://math-ci.amd.com/job/main/job/precheckin-cuda/job/hipBLAS/job/PR-889/2/pipeline-console/?selected-node=208), may need help getting hipblas-common as a dependency there as well.

rmake.py
```diff
@@ -318,6 +318,10 @@ def config_cmd():
         raw_rocsolver_path = cmake_path(args.rocsolver_path)
         rocsolver_path_cmake =  f'"{raw_rocsolver_path}"'
 
+    raw_hipblas_common_path = cmake_path("C:/hipSDK")
+    hipblas_common_path = f'"{raw_hipblas_common_path}"'
```
daineAMD: Yeah I guess would be good to add --hipblas_common_path cmd line argument. I can add it later though, doesn't have to hold up this PR.

library/src/CMakeLists.txt
```diff
@@ -67,7 +67,7 @@ add_library( roc::hipblas ALIAS hipblas )
 set(static_depends)
 
 if( WIN32 )
-  find_package( hipblas-common REQUIRED CONFIG PATHS ${HIPBLAS-COMMON_PATH})
+  find_package( hipblas-common REQUIRED CONFIG PATHS "C:/hipSDK")
```
TorreZuk: The C:/hipSDK should be changed to just ${ROCM_PATH} which should work on both WIN32 and linux

clients/CMakeLists.txt
```diff
@@ -82,6 +82,12 @@ list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake )
 
 include( build-options )
 
+if( WIN32 )
+  find_package( hipblas-common REQUIRED CONFIG PATHS "C:/hipSDK")
+else()
+  find_package( hipblas-common REQUIRED CONFIG PATHS /opt/rocm)
+endif()
```
TorreZuk: Why would the clients find this package?   hipBLAS should capture include paths for its dependencies I think is the way this should go.

library/src/CMakeLists.txt
```diff
@@ -66,6 +66,12 @@ add_library( roc::hipblas ALIAS hipblas )
 
 set(static_depends)
 
+if( WIN32 )
+  find_package( hipblas-common REQUIRED CONFIG PATHS "C:/hipSDK")
+else()
+  find_package( hipblas-common REQUIRED CONFIG PATHS /opt/rocm)
+endif()
```
TorreZuk: should remove hard coded C:/hipSDK but better to change both as per my earlier comment.

daineAMD:
Static CI looks like it's broken (for a long time). I'll see if I can test it out manually and approve afterwards.

daineAMD (approved):
Built static manually (without rocsolver) and seems to work ok. CI will be the real test but this should be fine and we can hotfix if needed.

PR #898, opened 2024-08-06
Develop branch for release after ROCm 6.3

daineAMD:
- hipBLAS version 2.3.0 -> 2.4.0
- rocBLAS dependency version 4.3.0 -> 4.4.0
- rocSOLVER dependency version 3.27.0 -> 3.28.0

Awaiting confirmation of rocSOLVER version number, will leave as a draft and noCI until rocSOLVER is ready. Also removed old bump_master_version.sh since it isn't used anymore.

release-staging/rocm-rel-6.3 has been created from develop with no version changes.

PR #900, opened 2024-08-07
Using BUILD_SOLVER_TESTS cmake variable in prep of runtime solver

daineAMD:
This change is in preparation for adding the feature of finding rocSOLVER at runtime using dlopen.

This change adds a new cmake option BUILD_SOLVER_TESTS to enable rocSOLVER tests. Currently in our install scripts, this simply follows BUILD_WITH_SOLVER which is disabled by the --no-solver flag. In a future commit, though, BUILD_WITH_SOLVER will be OFF by default, and will only specify whether solver should be found at buildtime. BUILD_SOLVER_TESTS will be ON by default, and the --no-solver flag will turn these tests off.

Will wait for #898 to go through to get new versioning, and will run CI with solver and w/o solver.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,6 +73,8 @@ option( BUILD_VERBOSE "Output additional build information" OFF )
 
 option( BUILD_WITH_SOLVER "Add additional functions from rocSOLVER" ON )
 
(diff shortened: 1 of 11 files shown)


daineAMD:
Closing in favour of a full PR including these changes.

Issue #896, opened 2024-07-31
[Feature]: Optional rocSOLVER at run time


cgmb:
### Suggestion Description

There is a `BUILD_WITH_SOLVER` CMake option, but it would be nice if the rocsolver library could be loaded at run-time with dlopen when `BUILD_WITH_SOLVER` is `OFF`. This would allow Debian and Ubuntu systems packages to build with `-DBUILD_WITH_SOLVER=OFF` and mark rocsolver as a recommended package, rather than a required dependency. The behaviour for `BUILD_WITH_SOLVER=ON` would be unchanged (and would be needed for static hipBLAS builds).

Users that wish to use hipBLAS for applications such as llama-cpp currently must install hipBLAS, which depends on rocSOLVER, which depends on rocSPARSE. Despite only wanting to run a few GEMMS, they end up installing several gigabytes of other libraries that provide functionality that they do not need for their particular application. The rocSOLVER library has made rocSPARSE an optional dependency for this reason, and I think it would be a good idea for hipBLAS to do the same.

The rocSOLVER library would still be installed with hipblas by default, but users could specifically request `--no-install-recommends` to avoid it (or could explicitly uninstall rocsolver).

### Operating System

Ubuntu

### ROCm Component

hipBLAS

ammallya:
Imported to ROCm/rocm-libraries

cgmb:
@ammallya, it would be nice to close these issues as duplicates of the new issues, rather than closing them as completed (as they are not completed).

PR #302, opened 2021-01-27
Performance script fixes

daineAMD:
It appears that rocm_smi.py has changed in a recent release, which broke some of the data-gathering in the script. These have been fixed/worked-around here.
Also fixed some bugs with the --cuda and --compare options.

Going to play around with the script a little more to make sure everything is working as expected.

PR #903, opened 2024-08-19
Load rocSOLVER with dlopen at runtime

daineAMD:
Don't have access to a windows machine so this hasn't been tested on windows yet.

See https://github.com/ROCm/hipBLAS/issues/896 for the request. Similar changes for hipSOLVER have been done at [hipSOLVER #302](https://github.com/ROCm/hipSOLVER/pull/302)

This change allows hipBLAS to build without rocSOLVER available at buildtime. By default, hipBLAS will try to find rocSOLVER at runtime using dlopen(). If rocSOLVER cannot be found, the rocsolver-based functions will return HIPBLAS_STATUS_NOT_IMPLEMENTED.

The `-n` flags has been changed to mean "don't build rocSOLVER tests".
The `--solver-buildtime` flags has been added to find rocSOVLER at build time.

Also added some flags for jenkins so will be testing the various flags on CI.


CMakeLists.txt
```diff
@@ -71,15 +71,19 @@ rocm_setup_version( VERSION ${VERSION_STRING} )
...
-if( BUILD_WITH_SOLVER )
-    add_definitions( -D__HIP_PLATFORM_SOLVER__ )
-endif( )
-
 # BUILD_SHARED_LIBS is a cmake built-in; we make it an explicit option such that it shows in cmake-gui
 option( BUILD_SHARED_LIBS "Build hipBLAS as a shared library" ON )
 
+if(BUILD_SHARED_LIBS)
+    set(BUILD_WITH_SOLVER_DEFAULT OFF)
+else()
```
TorreZuk, line 79: Is this just for testing or default to not link?    Default to not link may may need to review effect on rpath or package consumption documentation.

TorreZuk (commented):
Is the static build without solver_buildtime the default as that creates an invalid static library?

cgmb:
> Is the static build without solver_buildtime the default as that creates an invalid static library?

It's valid. It's just unlikely that you'd want a static hipblas and a dynamic rocsolver.

library/src/amd_detail/dlopen/load_rocsolver.cpp
```diff
@@ -0,0 +1,217 @@
...
+fp_rocsolver_zgels_batched         g_rocsolver_zgels_batched;
+fp_rocsolver_sgels_strided_batched g_rocsolver_sgels_strided_batched;
+fp_rocsolver_dgels_strided_batched g_rocsolver_dgels_strided_batched;
+fp_rocsolver_cgels_strided_batched g_rocsolver_cgels_strided_batched;
+fp_rocsolver_zgels_strided_batched g_rocsolver_zgels_strided_batched;
+
+static bool load_rocsolver()
+{
+#ifdef WIN32
+    void* handle = LoadLibraryW(L"librocsolver.dll");
```
cgmb: I believe it's just `rocsolver.dll`.

library/src/amd_detail/dlopen/load_rocsolver.cpp
```diff
@@ -0,0 +1,217 @@
...
+    return true;
+}
+
+#undef LOAD_FN
+
+#endif // BUILD_WITH_SOLVER
+
+bool try_load_rocsolver()
+{
+#ifndef __HIP_PLATFORM_SOLVER__
```
cgmb: I think you renamed this elsewhere.

library/src/amd_detail/dlopen/load_rocsolver.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+static bool load_rocsolver()
+{
+#ifdef WIN32
+    void* handle = LoadLibraryW(L"librocsolver.dll");
+#else
+
+    // RTLD_NOW to load symbols now to avoid performance hit later
+    // RTLD_GLOBAL to load geqrf_ptr_batched declared symbols in hipblas.cpp
+    void* handle = dlopen("librocsolver.so", RTLD_NOW | RTLD_GLOBAL);
```
cgmb: Please load `librocsolver.so.0`. `librocsolver.so` is only provided by the -dev/-devel package on Debian, Ubuntu and Fedora. If you want to fall back to `librocsolver.so` in the case that loading `librocsolver.so.0` fails, that's fine.

The unsuffixed symlink goes in the -dev package on those distros because the maintainers wish to ensure that both `librocsolver.so.0` and the hypothetical future `librocsolver.so.1` can both be installed at the same time without having a conflict caused by both packages containing a file with the same name. The ability to install the two versions side-by-side makes for a smoother migration during the package update process.

cgmb:
You may want to adjust the CMake that defines the package dependencies to recommend and/or suggest the rocsolver package to users that install hipblas.

TorreZuk:
> > Is the static build without solver_buildtime the default as that creates an invalid static library?
> 
> It's valid. It's just unlikely that you'd want a static hipblas and a dynamic rocsolver.

Invalid in the sense that neither a customer nor developer would want it (we don't want failing tests for bad configs), let alone be default.   Maybe a conditional option should be used?

CMakeLists.txt
daineAMD: Default will be to not link. I think it'll be ok but I'll test out consuming the hipblas package now to make sure it find rocsolver.

library/src/amd_detail/dlopen/load_rocsolver.cpp
daineAMD: Testing this out on a windows machine now.

cgmb:
> Invalid in the sense that neither a customer nor developer would want it (we don't want failing tests for bad configs), let alone be default. Maybe a conditional option should be used?

Keep in mind that we're not distinguishing between 'maybe rocsolver at runtime' and 'no rocsolver'. The configuration of 'static hipblas with no rocsolver' is something that developers may want.

TorreZuk:
> > Invalid in the sense that neither a customer nor developer would want it (we don't want failing tests for bad configs), let alone be default. Maybe a conditional option should be used?
> 
> Keep in mind that we're not distinguishing between 'maybe rocsolver at runtime' and 'no rocsolver'. The configuration of 'static hipblas with no rocsolver' is something that developers may want.

The flags are confusing me maybe keeping -n --no-solver as never any solver is simpler (both shared and static).  What this seems to be is an architeture option to  delays .so loading to make it optional.   Should discuss an overall roadmap feature request for this change for hip libraries.

library/src/CMakeLists.txt
```diff
@@ -82,6 +82,11 @@ if(HIP_PLATFORM STREQUAL amd)
     endif( )
   endif( )
 
+  if(NOT BUILD_WITH_SOLVER)
+    target_link_libraries(hipblas PRIVATE $<$<PLATFORM_ID:Linux>:${CMAKE_DL_LIBS}>)
+    set_property(TARGET hipblas APPEND PROPERTY BUILD_RPATH ${ROCSOLVER_PATH}/lib)
```
cgmb, line 87: What prompted this change?

daineAMD: I believe this is needed to find the rocsolver library if hipblas is built with the install script --rocsolver-path option

cgmb: Keep in mind that rocsolver might not actually even exist on the system at build-time. Personally, I wouldn't bother mucking around with the `RPATH`, as it can always be handled after the build by setting the `LD_LIBRARY_PATH`.

If you prefer to set the RPATH, it should be guarded to only be set when rocsolver exists and `--rocsolver-path` has been used. The hardcoded `lib` may be a problem, as the library directory can sometimes have other names in third-party builds, like `lib64` or `<target-triple>/lib`. You could use `${CMAKE_INSTALL_LIBDIR}` instead, but I think it would be more robust to extract the path from the target:
```cmake
if(DEFINED ROCSOLVER_PATH)
  find_package(rocsolver QUIET PATHS "${ROCSOLVER_PATH}" NO_DEFAULT_PATH)
  if(rocsolver_FOUND)
    get_target_property(rocsolver_location roc::rocsolver LOCATION)
    get_filename_component(rocsolver_directory ${rocsolver_location} DIRECTORY)
    set_property(TARGET hipblas APPEND PROPERTY BUILD_RPATH ${rocsolver_directory})
  endif()
endif()
```

With all that said, I prefer less code to more code. IMO, it's fine to expect users to set `LD_LIBRARY_PATH` if rocSOLVER is not in a system directory or `/opt/rocm`.

ammallya:
Unable to import due to merge conflict please reopen this PR on ROCm/rocm-libraries

PR #904, opened 2024-08-19
Fix hipBLAS-common packaging call

awhittle3:
resolves export packaging issues with hipBLAS-common 

Summary of proposed changes:
- Properly invoke `rocm_export_targets` on hipBLAS-common package
- Minor whitespace cleanup

### DIFF
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -184,20 +184,20 @@ rocm_install_targets(
 )
 #         PERMISSIONS OWNER_EXECUTE OWNER_WRITE OWNER_READ GROUP_EXECUTE GROUP_READ WORLD_EXECUTE WORLD_READ
 
-if(HIP_PLATFORM STREQUAL amd)
-    rocm_export_targets(
-        TARGETS roc::hipblas
-	DEPENDS PACKAGE hip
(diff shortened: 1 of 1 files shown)


bnemanich (approved):

library/src/CMakeLists.txt
```diff
@@ -184,20 +184,20 @@ rocm_install_targets(
...
 else( )
-    rocm_export_targets(
-        TARGETS roc::hipblas
-	DEPENDS PACKAGE HIP
-	NAMESPACE roc::
-    )
+  rocm_export_targets(
+    TARGETS roc::hipblas
+    DEPENDS PACKAGE HIP
+    NAMESPACE roc::
```
TorreZuk, line 199: Even the cuda backend 
DEPENDS PACKAGE hipblas-common 
as the header is included at top level.

TorreZuk (approved):
Is there a packaging reason why we do't have a line like:
list(APPEND static_depends PACKAGE hipblas-common)
Maybe no static package?   Fine if you test and working as expected as no CI tests for this I think so you need manual test.

library/src/CMakeLists.txt
awhittle3: Requested change as part of new PR #906

PR #920, opened 2024-09-25
Update dependency  list with the actual package name

raramakr:
hipblas-common-devel/dev is the actual package name and it provides hipbals-common. Use the actual package name in dependency.

resolves #___

Summary of proposed changes:
-
-
-


TorreZuk (approved):
LGTM

TorreZuk:
This would also be required cherry pick @daineAMD , right?

TorreZuk:
@daineAMD can you review and okay merge?

daineAMD:
hipBLAS will need another cherry-pick into 6.3 anyway so I can include this there too.

PR #921, opened 2024-09-26
Dependency on hipblas-common-dev

awhittle3:
resolves #SWDEV-487853

Summary of proposed changes:

The dependency on [hipblas-common](https://github.com/ROCm/hipBLAS-common) actually needs to point to the hipblas-common-dev package.

With this change, hipblas-dev package will now list hipblas-common-dev as a dependency.

Thanks to @lawruble13 for the fix!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,8 +212,8 @@ if(HIP_PLATFORM STREQUAL amd)
 endif( )
 
 set(hipblas_common_minimum 1.0.0)
-rocm_package_add_rpm_dependencies(DEPENDS "hipblas-common-devel >= ${hipblas_common_minimum}")
-rocm_package_add_deb_dependencies(DEPENDS "hipblas-common-dev >= ${hipblas_common_minimum}")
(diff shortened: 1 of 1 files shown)


TorreZuk:
@awhittle3 this change mostly already went in, but if the COMPONENT is required please rebase

bnemanich (approved):

awhittle3:
> @awhittle3 this change mostly already went in, but if the COMPONENT is required please rebase

The change from #920 is not incorrect, but the base `hipblas` package probably doesn't need the headers that `hipblas-common-dev` provides? Rebased as requested.

TorreZuk (approved):
This is the same name fix it just sets it for COMPONENT devel rather than without a component, which not clear would have set it for both?

lawruble13:
Setting up the dependency without a component would normally set it for both, however because the devel component has an implicit dependency on the runtime component, it actually ends up using the `DEVEL` version of the variable, which means that it actually won't use the values in the unqualified dependency list. This behaviour was intended back in the early days of the split in packaging, so that there would be a minimum amount of code change needed.

PR #930, opened 2024-10-11
add graph support documentation

amcamd:


docs/functions.rst
```diff
@@ -111,6 +111,11 @@ Some functions in hipBLAS may use atomic operations to increase performance whic
 By default, the rocBLAS backend allows the use of atomics while the cuBLAS backend disallows the use of atomics. To set the desired behavior, users should call
 :any:`hipblasSetAtomicsMode`. Please see the rocBLAS or cuBLAS documentation for more information regarding specifics of atomic operations in the backend library.
 
+Graph Support for hipBLAS
+=========================
+
+Graph support for hipBLAS is the same as that in the rocBLAS or cuBLAS backend. If rocBLAS is the backend, see rocBLAS documentation. If cuBLAS is the backend, see cuBLAS documentation.
```
TorreZuk: This sounds good.  But what is graph support?   :-)    I guess we don't have a hip reference we can link to.

TorreZuk (approved):
Fine as is to me

rkamd (approved):
LGTM. 
Minor suggestion, we could reword it as below:

Graph support ( also referred to stream capture support) for hipBLAS depends on the backend in use. if rocBLAS is the backend, refer to rocBLAS documentation. Similarly, if cuBLAS is the backend, refer to cuBLAS documentation.

amcamd:
Thanks for comments, changed to suggested text.

PR #944, opened 2024-12-17
Use jenkins xml over standard preset node config

amd-garydeng:
Math-CI wants to be able to set node configs for static analysis job instead of the default ubuntu20:['any']


TorreZuk (approved):
LGTM

samjwu (approved):

PR #959, opened 2025-02-03
add hipblasSetWorkspace

jeffdaily:


jeffdaily:
Not sure how to write a unit test for this one since there is no corresponding query AFAICT.  pytorch started using cublasSetWorkspace() and hipblas didn't have it even though rocblas did.  I put in an ugly hack into pytorch to call rocblas directly.  See https://github.com/pytorch/pytorch/pull/138791.

jeffdaily:
Should we close this in favor of #960?

TorreZuk:
> Should we close this in favor of #960?

Yes I think so as @daineAMD can preserve your commit or add co_author so think you can close this. Best to merge new code along with new test coverage.

amcamd:
Yes, @jeffdaily @daineAMD go ahead and close this one and merge https://github.com/ROCm/hipBLAS/pull/960 . As long as https://github.com/ROCm/hipBLAS/pull/960 is `merged` and not `squash and merge` both commits will be preserved.

PR #950, opened 2025-01-14
Refactor hipBLAS documentation

amd-jnovotny:
resolves #___

Summary of proposed changes:
- Refactor hipBLAS documentation in accordance with style guide and standard formatting
- Copy edits
- Reorganize directory structure to match docs layout (for example, create how-to directory and move a file there)

PR #962, opened 2025-02-13
Refactor hipBLAS documentation (#950)

amcamd:
**Need to make sure this only contains 6.4 content**

* Initial draft. Requires more work before push

* Refactor remaining files and additional edits

* Apply suggestions from team review

* Fix errors in API reference

* Incorporate comments from review

* A few additional edits


amd-jnovotny (approved):
I did some spot checks, and it looks all right. I think the docs folder is the same between 6.4 and develop with the exception of dependabot/rocm-docs-core changes, which shouldn't affect this PR.

TorreZuk (approved):
I think it looks to be all there.  Maybe get Daine to approve as well.   Has the deprecation always been in oldest to newest order?   I think newest to oldest would be an improvement.

amd-jnovotny:
@TorreZuk : Yes, hipBLAS deprecations were listed oldest to newest. I just fixed some formatting problems and typos. But your idea is a good one. I could potentially change this with a new PR if you'd like. I have to handle some higher priority work right now but could circle back to this.

TorreZuk:
> @TorreZuk : Yes, hipBLAS deprecations were listed oldest to newest. I just fixed some formatting problems and typos. But your idea is a good one. I could potentially change this with a new PR if you'd like. I have to handle some higher priority work right now but could circle back to this.

Fine to reorder only into develop for next release.  Is this consistent across libraries?

amd-jnovotny:
@TorreZuk As it turns out, the order for rocBLAS was mixed up and in no consistent order, so I put that one into newest to oldest. So definitely it would be good to bring hipBLAS into line with that.

Not every repo has a deprecation listing and some just have it at the bottom of their API docs. I'll do a survey to see if there's a pattern.

daineAMD (approved):

PR #968, opened 2025-02-27
Static dep for hipblas-common

daineAMD:
resolves #___

Summary of proposed changes:
-
-
-


TorreZuk (approved):
LGTM if build passes.  I know the name went back and forth so CI vs official build.   I didn't hear that it was failing in mainline....

PR #974, opened 2025-03-11
Find gtest in clients CMakeLists

daineAMD:
Shouldn't need gtest_main from GTEST_BOTH_LIBRARIES I don't think, but this matches rocblas.

### DIFF
--- a/clients/CMakeLists.txt
+++ b/clients/CMakeLists.txt
@@ -195,6 +195,11 @@ if( BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
     set( BLIS_CPP ../common/blis_interface.cpp )
   endif()
 
+  if(EXISTS  "${BUILD_DIR}/deps/deps-install/lib/libgtest.a")
+    set( GTEST_ROOT "${BUILD_DIR}/deps/deps-install")
+  endif()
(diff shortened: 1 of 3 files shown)


clients/CMakeLists.txt
```diff
@@ -195,6 +195,11 @@ if( BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
     set( BLIS_CPP ../common/blis_interface.cpp )
   endif()
 
+  if(EXISTS  "${BUILD_DIR}/deps/deps-install/lib/libgtest.a")
+    set( GTEST_ROOT "${BUILD_DIR}/deps/deps-install")
+  endif()
+  find_package( GTest REQUIRED )
```
TorreZuk, line 201: Could wrap in if build_client_benchmarks or build_client_tests?   If supporting partial building of clients is of interest....

TorreZuk (approved):
Suggestion only, LGTM

clients/CMakeLists.txt
daineAMD: Already wrapped in `if( BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)` from L112 - L214

PR #978, opened 2025-03-24
Update hipblas-common version for ROCm 6.5

daineAMD:
Don't necessarily need the updated hipblas-common version, but matches our pattern of requesting versions from same rocm release as rocblas/rocsolver.


TorreZuk (approved):
Fine by me

PR #979, opened 2025-03-31
add codecov config

TorreZuk:
* trial run


daineAMD (approved):
Does the codecoverage report display correctly for you? I just see a bunch of EOFs when I look at the individual files.

TorreZuk:
> Does the codecoverage report display correctly for you? I just see a bunch of EOFs when I look at the individual files.

Okay function listed correctly but source parsing must be bad, rocBLAS source showing correctly so will compare

PR #980, opened 2025-03-31
Example of deprecation removals

daineAMD:
Opening this PR to see if there's any comments/concerns regarding the removal of deprecated hipblas functions/types. See comment in [LWPMLSE-820](https://ontrack-internal.amd.com/browse/LWPMLSE-820) for some discussion regarding method of removal of deprecations.

This PR shows what the removal will look like from the library side for a single function, along with the removal of `hipblasComplex`, `hipblasDoubleComplex`, and `hipblasDatatype` types. Further to this PR, client-side changes will be needed to use hipComplex/hipDoubleComplex directly. This PR will fail CI as I've only shown what the changes look like for `hipblasXcopy()`.

If this method is fine with the team I'll put together a PR with more changes, potentially I'll do all the client-side changes so I can make piecewise changes on the library to make reviewing easier.


TorreZuk (approved):
That looks good to me.   A lot of code removed so will be nice.

daineAMD:
See #982 for full implementation.

PR #983, opened 2025-04-09
move to newer package name and fallback

TorreZuk:
* resolves deprecated use

### DIFF
--- a/cmake/get-rocm-cmake.cmake
+++ b/cmake/get-rocm-cmake.cmake
@@ -8,7 +8,7 @@ if(NOT ROCmCMakeBuildTools_FOUND)
   find_package(ROCM 0.11.0 CONFIG QUIET PATHS "${ROCM_PATH}") # deprecated fallback
   if(NOT ROCM_FOUND)
     message(STATUS "ROCmCMakeBuildTools not found. Fetching...")
-    set(rocm_cmake_tag "rocm-6.3.3" CACHE STRING "rocm-cmake tag to download")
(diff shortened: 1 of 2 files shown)


cmake/get-rocm-cmake.cmake
```diff
@@ -0,0 +1,21 @@
...
+# rocm-cmake contains common cmake code for rocm projects to help setup and install
+
+include(FetchContent)
+
+find_package(ROCmCMakeBuildTools 0.11.0 CONFIG QUIET PATHS "${ROCM_PATH}")
+if(NOT ROCmCMakeBuildTools_FOUND)
+  find_package(ROCM 0.11.0 CONFIG QUIET PATHS "${ROCM_PATH}") # deprecated fallback
+  if(NOT ROCM_FOUND)
+    message(STATUS "ROCmCMakeBuildTools not found. Fetching...")
+    set(rocm_cmake_tag "rocm-6.3.3" CACHE STRING "rocm-cmake tag to download")
```
daineAMD: 6.3.3 until 6.4 is released?

daineAMD (approved):
LGTM but would be good to get a rocm-cmake expert review as well.

cmake/get-rocm-cmake.cmake
TorreZuk: > 6.3.3 until 6.4 is released?

Sure but probably fine to leave on older release unless new helper function required

TorreZuk:
@lawruble13 if you can review today let me know otherwise I'll merge and can accept feedback later

PR #984, opened 2025-04-09
Reorder hipBLAS deprecation notices with newest on top

amd-jnovotny:
resolves #___

Summary of proposed changes:
- Reorders the deprecation notices from newest to oldest.
- No copy edits of the actual notices and no other changes.
-


TorreZuk (approved):
Agree better ordering

PR #989, opened 2025-04-11
Work around race condition

benrichard-amd:
When building the hipBLAS clients, Fortran modules are compiled. When using a machine with enough parallelism due to a high core count, compiling the hipblas_fortran_client target fails with

> hipblas.mod’ for reading at (1): No such file or directory"

It succeeds in subsequent runs as hipblas.mod is created in the first run but not timely enough.

One possibility to resolve this is to replace add_dependency() with target_link_libraries() thus that hipblas.mod is present when needed.

Note: This is part of a set of patches required to get TheRock (https://github.com/ROCm/TheRock) to build properly.


TorreZuk:
Can you confirm that the .mod file is not being generated in the wrong location in the first run when it fails?  I think that is a different issue than add_dependencies not requiring a dependency to be built first.

daineAMD:
Yeah I recall the .mod file being created in one of the library/ directories at one point, which causes issues on rebuilds of the module if there are changes. Not sure if this was ever fixed.

TorreZuk:
Yes if you can consistently reproduce the .mod creation glitch better to fix that than just tweak dependency.   I think it may be just the fortran cmake variables aren't set early enough, if I could ever reproduce the output failure consistently I was going to fix it.   @benrichard-amd  were the .mod files merged by the link, or cmake just propagated the link argument.  Can you paste the relevant make file code and linker lines to be clear.

stellaraccident:
@marbre who originally authored this patch

marbre:
When building with hipBLAS with TheRock
```
[9/192] Building Fortran object clients/CMakeFiles/hipblas_fortran_client.dir/include/hipblas_fortran_no_solver_module.f90.o
FAILED: clients/CMakeFiles/hipblas_fortran_client.dir/include/hipblas_fortran_no_solver_module.f90.o clients/hipblas_interface.mod 
/usr/bin/gfortran -I/TheRock/math-libs/BLAS/hipBLAS/clients/include -I/TheRock/build/math-libs/BLAS/hipBLAS/build/include/hipblas -I/TheRock/
build/math-libs/BLAS/hipBLAS/build/include -O3 -Jclients -fpreprocessed -c clients/CMakeFiles/hipblas_fortran_client.dir/include/hipblas_fortran_no_solver_module.f90-pp.f90 -o clients/CMakeFi
les/hipblas_fortran_client.dir/include/hipblas_fortran_no_solver_module.f90.o
/TheRock/math-libs/BLAS/hipBLAS/clients/include/hipblas_fortran_no_solver_module.f90:26:9:

   26 |     use hipblas
      |         1
Fatal Error: Cannot open module file ‘hipblas.mod’ for reading at (1): No such file or directory
compilation terminated.
```
can be consistently reproduced with the following:

```bash
# Checkout sources 
git clone git@github.com:ROCm/TheRock.git
cd TheRock

# Install Python dependencies, see https://github.com/ROCm/TheRock?tab=readme-ov-file#common
python3 -m venv venv
source venv/bin/activate
pip install -r requirements.txt

# Install system dependencies, see https://github.com/ROCm/TheRock?tab=readme-ov-file#on-ubuntu
sudo apt install gfortran git-lfs ninja-build cmake g++ pkg-config xxd patchelf automake


# Delete patch (!), fetch sources and apply remaining patches
rm patches/amd-mainline/hipBLAS/0002-Work-around-race-condition.patch
./build_tools/fetch_sources.py

# Configure the build
cmake -B build -GNinja . -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_C_COMPILER_LAUNCHER=ccache -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DTHEROCK_AMDGPU_FAMILIES=gfx110X-dgpu -DTHEROCK_PACKAGE_VERSION=ADHOCBUILD -DTHEROCK_VERBOSE=ON -DBUILD_TESTING=ON -DTHEROCK_ENABLE_ALL=OFF -DTHEROCK_ENABLE_BLAS=ON

# Try to build hipBLASD
cmake --build build --target hipBLAS
```

With the patch restored and applied via `git checkout patches/amd-mainline/hipBLAS/0002-Work-around-race-condition.patch && ./build_tools/fetch_sources.py`, hipBLAS can be successfully build.


As in the issue / patch description (patch tracked in https://github.com/ROCm/TheRock/issues/260), replacing `add_dependency()` with `target_link_libraries()` is probably only _one_ option. I tried to move `CMAKE_Fortran_MODULE_DIRECTORY` but was not able to resolve it, ending up with the patch under review.

TorreZuk:
@marbre where did the .mod file get created?   Ninja I have experience bugs with incorrect dependencies, so are you making tickets for Ninja?  If you trace your build you may spot the correct fix or be able to file a ticket.

stellaraccident:
> @marbre where did the .mod file get created? Ninja I have experience bugs with incorrect dependencies, so are you making tickets for Ninja? If you trace your build you may spot the correct fix or be able to file a ticket.

Ninja never had incorrect dependency bugs in all of my years using it. However it often exposes bugs in projects like this which have been using a less aggressive build system for too long. We just have to fix it.

I would need to study this more closely in order to root cause: cmake actually has a subtle difference between file and target dependencies and it is a frequent cause of these kinds of issues because adding a dependency on a target that produces a file is not always sufficient to ensure the file is built before. Iirc, target_link_libraries also has the side effect of establishing a file level dependency which could be an explanation for the behavior difference. I have to debug one of this class of issues every year or so and I may be remembering incorrectly and need to refresh cache.

TorreZuk:
I agree with you I would first follow the lead of CMAKE_Fortran_MODULE_DIRECTORY, that is why I originally asked where is the .mod file.  Probably not a ninja bug.

AI supports my memory, as reports the Ninja build system has encountered various dependency-related bugs over time, as documented in its release logs and issue tracker. Here are a few notable examples:

Phony Targets and Restat Issues: A bug was identified where the mtime for outputs of rules with restat = 1 and phony targets was incorrectly set to 0. This caused builds to be considered "dirty" repeatedly.

Large Dependency Files: Ninja misbehaved when handling files with an exceptionally large number of dependencies, leading to crashes or errors like "premature end of file." This issue required changes to the binary format.

Missing Dependency Graph Edges: In large projects, missing dependencies between targets could lead to build failures. Tools like ninja -t deps and ninja -t graph were suggested for diagnosing these issues.

stellaraccident:
> I agree with you I would first follow the lead of CMAKE_Fortran_MODULE_DIRECTORY, that is why I originally asked where is the .mod file. Probably not a ninja bug.
> 
> AI supports my memory, as reports the Ninja build system has encountered various dependency-related bugs over time, as documented in its release logs and issue tracker. Here are a few notable examples:
> 
> Phony Targets and Restat Issues: A bug was identified where the mtime for outputs of rules with restat = 1 and phony targets was incorrectly set to 0. This caused builds to be considered "dirty" repeatedly.
> 
> Large Dependency Files: Ninja misbehaved when handling files with an exceptionally large number of dependencies, leading to crashes or errors like "premature end of file." This issue required changes to the binary format.
> 
> Missing Dependency Graph Edges: In large projects, missing dependencies between targets could lead to build failures. Tools like ninja -t deps and ninja -t graph were suggested for diagnosing these issues.

Far be it from me to argue with AI. I have definitely seen and fixed ninja bugs, but it has been a very long time / never since I have seen ninja not enforce the dependency chain and proceed to do the wrong thing by invoking two actions in the wrong order -- which is what this class of issue would be. Burden of proof for that is in the project, and ime is caused almost all of the time by one of these subtle cmake constraints that you need to read carefully (and within that subset almost always by the target/file dep dichotomy I mentioned).

There's always a first time for an outright ninja correctness bug but this is probably not it. In any case, this one should be reduced to a more satisfying root cause.

marbre:
> @marbre where did the .mod file get created?

`.mod` files get created at

> build/math-libs/BLAS/hipBLAS/build/clients/hipblas_interface.mod
build/math-libs/BLAS/hipBLAS/build/include/hipblas.mod
build/math-libs/BLAS/hipBLAS/build/include/hipblas_enums.mod

TorreZuk:
Okay the .mod look to be where I expected them.  So for more AI advice, there are known issues with Ninja's handling of .mod files in Fortran projects, particularly when used with CMake. These bugs often arise due to incorrect dependency resolution between Fortran modules during parallel builds. For example:

Build Order Problems: Ninja may attempt to compile a module before its dependencies are ready, leading to errors like "Cannot open module file."

CMake Version-Specific Issues: Some versions of CMake (e.g., 3.27.0) have been reported to generate faulty Ninja build files for Fortran projects, causing frequent build errors2.

Workarounds include:

Using an older version of CMake (e.g., 3.26.x) that handles Fortran dependencies more reliably.

Switching to a different build system, such as Unix Makefiles, which may avoid these issues altogether.

Manually enforcing build order using add_dependencies or specifying Fortran_MODULE_DIRECTORY in CMake.

target_link_libraries in CMake is not designed to work directly with Fortran .mod files. .mod files are module files generated by Fortran compilers and are used during compilation, not linking. Instead, you should:

Specify Include Directories: Use target_include_directories to include the directory where the .mod files are located:

cmake
target_include_directories(your_target PRIVATE ${CMAKE_Fortran_MODULE_DIRECTORY})
Set the Module Directory: Ensure that the CMAKE_Fortran_MODULE_DIRECTORY variable is set to the directory where .mod files will be generated:

cmake
set(CMAKE_Fortran_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/modules)
Link Libraries: Use target_link_libraries for linking actual library files (e.g., .a, .so, or .lib), not .mod files.  

The linker error in Azure seems to indicate you converted it to a .so which we do not want here.

stellaraccident:
Pretty sure this is the usual problem: you need to depend on the file, not the producing target. target_link_libraries does this implicitly, whereas add_dependencies only takes a target level dep.

(Ie. The reason why trying to link it as a library is producing a proper order is because of the byproduct of it taking the dependency on the actual files it needs)

ammallya:

PR #992, opened 2025-04-14
Replace refs to GTEST_BOTH_LIBRARIES with individual libraries.

### LINKED ISSUE #413: [hipBLAS] Replace refs to GTEST_BOTH_LIBRARIES with individual libraries.
Upstream: https://github.com/ROCm/hipBLAS/pull/992

stellaraccident:
* This variable is actually only provided by the finder module and is obsolete (the upstream package does not provide it).
* Modern cmake norms just calls for using the libraries themselves.

Fixes https://github.com/ROCm/TheRock/issues/413

### DIFF
--- a/clients/benchmarks/CMakeLists.txt
+++ b/clients/benchmarks/CMakeLists.txt
@@ -82,8 +82,8 @@ target_include_directories( hipblas_v2-bench
     $<BUILD_INTERFACE:${FLAME_INCLUDE_DIR}>
 )
 
-target_link_libraries( hipblas-bench PRIVATE roc::hipblas ${GTEST_BOTH_LIBRARIES} )
-target_link_libraries( hipblas_v2-bench PRIVATE roc::hipblas ${GTEST_BOTH_LIBRARIES} )
+target_link_libraries( hipblas-bench PRIVATE roc::hipblas GTest::gtest GTest::gtest_main )
+target_link_libraries( hipblas_v2-bench PRIVATE roc::hipblas GTest::gtest GTest::gtest_main )
 
 if (NOT WIN32)
     target_link_libraries( hipblas-bench PRIVATE hipblas_fortran_client )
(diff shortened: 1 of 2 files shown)


TorreZuk (commented):
@stellaraccident did you test with a build from source of GTest in the deps build tree, what version of GTest are you using?  We actually had a target style dependency recently vs. the legacy variable.  @daineAMD what OS was failing with that do you recall that recent change?  I only recall way back the cmake gtest targets style failed on some OS but all supported the GTEST_BOTH_LIBRARIES ... maybe those OS have rolled off our supported list.   Backward compatibility is often better than using latest style.

stellaraccident:
> @stellaraccident did you test with a build from source of GTest in the deps build tree, what version of GTest are you using? We actually had a target style dependency recently vs. the legacy variable. @daineAMD what OS was failing with that do you recall that recent change? I only recall way back the cmake gtest targets style failed on some OS but all supported the GTEST_BOTH_LIBRARIES ... maybe those OS have rolled off our supported list. Backward compatibility is often better than using latest style.

I'm using cmake 3.25 with a gtest_DIR explicitly set to the package. Afaict, that can never work with these old variables because they were provided by the find module, not the actual package itself.

So its not just a style thing: how it is written effectively can't work with a bring your own package as far as I can tell. 

Between the gtest and cmake folks, they indeed did make a huge mess of this and have produced multiple situations where what is backwards compatible isn't forward compatible, etc :/

Some of the rocm projects include a shim that aliases targets, which is another way to ensure more backwards compatibility, but afaict, using these variables is strongly deprecated and the least supported option since they only are available in contexts that found the package via a find module.

I believe that the targets I've listed here are available in cmake>=3.20. If we need further backwards compatibility, we should crib the fragment that is in a few Dependency.cmake files through the project which aliases the old target names to the new.

daineAMD (approved):
Relevant: https://github.com/ROCm/hipBLAS/pull/974. I noted that we don't actually need gtest_main in that PR. We did have the GTest::GTest target before that change. I'm fine with this.

TorreZuk (approved):
Windows CI build succeeded (testing not running due to unrelated issues) so the vcpkg installed version is fine with this.  GTEST_BOTH_LIBRARIES was required for older builds from source on some OS as target names were tried that much I do remember.   We can assume those rolled off support so for now won't suggest if (NOT GTest::gtest) nor FindGTest and CMake minimum version.

PR #994, opened 2025-04-15
update the libgfortran pkg dependencies for almalinux

srekolam:
resolves #___ 

Summary of proposed changes:
- the PR updates  rpm package names that are currently relevant for almalinux builds.
-
-


CMakeLists.txt
```diff
@@ -149,15 +149,9 @@ if( BUILD_CLIENTS_SAMPLES OR BUILD_CLIENTS_TESTS OR BUILD_CLIENTS_BENCHMARKS )
...
-  if(CLIENTS_OS STREQUAL "centos" OR CLIENTS_OS STREQUAL "rhel")
-    if(CLIENTS_OS_VERSION VERSION_GREATER_EQUAL "8")
-      set(GFORTRAN_RPM "libgfortran")
-    endif()
-  elseif(CLIENTS_OS STREQUAL "ubuntu" AND CLIENTS_OS_VERSION VERSION_GREATER_EQUAL "20.04")
-    set(GFORTRAN_DEB "libgfortran5")
-  elseif(CLIENTS_OS STREQUAL "mariner" OR CLIENTS_OS STREQUAL "azurelinux")
+  set(GFORTRAN_RPM "libgfortran")
+  set(GFORTRAN_DEB "libgfortran5")
+  if(CLIENTS_OS STREQUAL "mariner" OR CLIENTS_OS STREQUAL "azurelinux")
```
TorreZuk: Okay are the old OS support you have clearly dropped documented somewhere higher up as not supported anymore or do we need to put some documentation in each repo?  Assume this will roll across all repos.  @amcamd @daineAMD

amcamd (approved):

CMakeLists.txt
pramenku: @TorreZuk 
You meant for CentOS and Ubuntu 20.04?
This PR is actually not removing these distros. 
Eventually, all RPM has libgfortran and all DEB has libgfortran5 which will work for CentOS and UB20.04

For azurelinux, we have seperate as for this distro, it is different.

pramenku: Pasting data for different distro. For SLES 15, this PR might break as SLES has libgfortran5 and libgfortran4

RHEL8
$ rpm -qa | grep -i libgfortran
**libgfortran**-8.5.0-26.el8_10.x86_64

RHEL9
]$  rpm -qa | grep -i libgfortran
**libgfortran**-11.5.0-5.el9_5.x86_64

UB20
$dpkg -l | grep -i libgfortran5
ii  **libgfortran5**:amd64      10.5.0-1ubuntu1~20.04        amd64        Runtime library for GNU Fortran applications


UB22
$ dpkg -l | grep -i libgfortran
ii  **libgfortran5**:amd64                      12.3.0-1ubuntu1~22.04                   amd64        Runtime library for GNU Fortran applications

SLES 15
$ rpm -qa | grep -i libgfortran
**libgfortran5**-14.2.0+git10526-150000.1.6.1.x86_64
**libgfortran4**-7.5.0+r278197-150000.4.44.1.x86_64

AZL3
$  rpm -qa | grep -i gfortran
gfortran-13.2.0-7.azl3.x86_64

pramenku: We can go generic with below
  set(GFORTRAN_RPM "libgfortran")
  set(GFORTRAN_DEB "libgfortran5")
 
For RHEL7/SLES 15, we can keep If condition to check for "libgfortran4" like we did for Azure Linux.

TorreZuk: Well for lower risk why not just keep the original default structure and just add the new OS into the the pattern.  I think this is what @cgmb was suggesting in one of the other PRs.   Changing a default, even when a bad default, often causes unintended breakages.

pramenku: > Well for lower risk why not just keep the original default structure and just add the new OS into the the pattern. I think this is what @cgmb was suggesting in one of the other PRs. Changing a default, even when a bad default, often causes unintended breakages.

@TorreZuk  What ever default earlier is kept is also not correct for majority of the OS.
Even what I have suggested will go with lower risk.  libgfortran4 is for SLES and RHEL7. RHEL7 is not more ROCM supported so we all should not worry for RH7. SLES is only what we are keeping libgfortran4 so always good to keep default for major supported OS and keep If condition for few distros like SLES is here.

TorreZuk: @pramenku it was tested and correct for most earlier distributions that is the risk I am talking about, there are Centos, Rocky, Scientific Linux ... that have rh7 basis etc that would have been libgfortran4.  The default should be rarely changed IMO.   So if you are saying your risk is that other new OS aren't in the new "set" for libgfortran5 why can't they be easily added?    If you want to roll off gfortran4 you need to add a documentation item for this change then I am fine with it.   Can also move to internal discussion to bring in more people.

TorreZuk (approved):
LGTM

TorreZuk:
Note all these PRs aren't going into develop as usual so we need to ensure early merge back.

CMakeLists.txt
pramenku: @TorreZuk 
>>The default should be rarely changed IMO

Below is kept default for DEB which is now changed to "libgfortran5" if you check on UB20/22/24. FYI. I have shared above already output of different distro
set(GFORTRAN_DEB "libgfortran4")

I am fine if we want to keep default for RPM as below.
set(GFORTRAN_RPM "libgfortran4")

srekolam:
can you please see if this can be merged.

PR #982, opened 2025-04-07
Remove hipblasComplex, hipblasDatatype and remove HIPBLAS_V2 API

daineAMD:
Again, sorry for the huge PR, but with the testing infrastructure I found it difficult to split this into smaller PRs.
Changes include:

- Client-side
    - remove hipblas_v2-test and hipblas_v2-bench along with infrastructure for this
    - use hipComplex directly for testing. Mostly renaming, but also needed to add/change a couple helpers
    - hipblasDatatype -> hipDatatype renaming
- Library-side
    - Remove all V2 APIs. Each function which previously contained a `hipblasDatatype` or `hipblasComplex` type had a V2 version
    - Change all APIs to match the V2 signature. I.e. change all `hipblasDatatype` -> `hipDatatype` and `hipblasComplex` -> `hipComplex`, along with `hipblasDatatype` -> `hipblasComputeType_t` for `gemm_ex`
    - Change Fortran API. This did not have a V2 version as mentioned in the initial deprecation

Any users who ignored the deprecation warning and have not been compiling with HIPBLAS_V2 defined will see breaking changes after this PR. If users have been using the HIPBLAS_V2 define, then they shouldn't see any changes.

Edit: I see cuda build is failing so might have a couple more changes coming.

PR #995, opened 2025-04-17
Use std::complex for clients

daineAMD:
Client-side changes in prep for removing hipblas_v2 API. Uses std::complex for everything client-side, and cast to hipblasComplex when calling hipblas API. Client-side changes should then be minimal when switching to hipComplex types.

### DIFF
--- a/clients/common/cblas_interface.cpp
+++ b/clients/common/cblas_interface.cpp
@@ -1,5 +1,5 @@
 /* ************************************************************************
- * Copyright (C) 2016-2024 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (C) 2016-2025 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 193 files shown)


TorreZuk (commented):
Probably will ask for internal_type explanation next week.  Cheers!

TorreZuk (approved):
Probably need to add in some downstream reviewers....

jeffdaily:
I don't know that I'm qualified to review this?  All these changes are made to "client" which I interpret as your testing or example code, not the public APIs.  Can someone explain to me in a call what these changes are accomplishing exactly?

TorreZuk:
> I don't know that I'm qualified to review this? All these changes are made to "client" which I interpret as your testing or example code, not the public APIs. Can someone explain to me in a call what these changes are accomplishing exactly?

@daineAMD can explain but this is more of just an early heads up for the V2 API becoming the only API.   You don't need to scan the code changes.

daineAMD:
@jeffdaily https://github.com/ROCm/hipBLAS/pull/982 may have more relevant code changes.
hipBLAS will be removing `hipblasComplex` types and `hipblasDataType` types to migrate to HIP types. The new API has been available via a HIPBLAS_V2 define for quite some time, but this will now be an API breaking change where all hipBLAS APIs will be migrated to use the new types. [Deprecation Notice](https://rocm.docs.amd.com/projects/hipBLAS/en/docs-6.4.0/reference/deprecation.html#announced-in-hipblas-2-0).
Don't need to review changes, but good to have downstream teams aware of the changes coming if they haven't migrated to the new hipBLAS API, as if they haven't yet then breaking changes will be coming very soon.

jeffdaily:
Thanks for the notice.  pytorch has been using HIPBLAS_V2 for quite some time now, so we're ready.  Not sure about tensorflow or other hipblas consumers.

PR #997, opened 2025-04-30
Docs: Fix broken links to hipblas-common

amd-jnovotny:
resolves #___

Summary of proposed changes:
- Fixed some broken links in reference/hipblas-api-functions.htm to hipblas-common. They were pointing to a page that did not exist.
- The links now point to https://github.com/ROCm/hipBLAS-common/blob/develop/library/include/hipblas-common/hipblas-common.h
-

### DIFF
--- a/docs/reference/hipblas-api-functions.rst
+++ b/docs/reference/hipblas-api-functions.rst
@@ -147,7 +147,7 @@ hipBLAS types
 *************
 
 For information about the ``hipblasStatus_t``, ``hipblasComputeType_t``, and ``hipblasOperation_t`` enumerations,
-see the `hipblas-common <https://github.com/ROCm/hipBLAS-common/blob/develop/library/include/hipblas-common.h>`_ GitHub repository.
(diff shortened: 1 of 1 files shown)


lpaoletti (approved):

TorreZuk:
https://github.com/ROCm/hipBLAS-common/pull/16 looks like it breaks all our old documemtation.  This maybe should be reverted.

amcamd (approved):
>

amd-jnovotny:
@TorreZuk I'm making the change to account for the new directory structure, but if it's reverted, I'll need to change back.

TorreZuk:
All the documentation for all previous releases would give 404 errors wouldn't it?

TorreZuk:
I would have held off until we know if those changes are required but fine for now.

davidd-amd (commented):
This seems like a fragile practice. Consider linking the project rather than the file.

amd-jnovotny:
@ellosel That is the new proposal, see https://github.com/ROCm/hipBLAS/pull/998. I'm waiting for feedback from the team on that.

PR #998, opened 2025-04-30
Docs: revised fix for link to hipblas-common GitHub

amd-jnovotny:
resolves #___

Summary of proposed changes:
- Second attempt to fix the links to hipblas-common. I am now just linking to the repository root (https://github.com/ROCm/hipBLAS-common)

### DIFF
--- a/docs/reference/hipblas-api-functions.rst
+++ b/docs/reference/hipblas-api-functions.rst
@@ -147,7 +147,7 @@ hipBLAS types
 *************
 
 For information about the ``hipblasStatus_t``, ``hipblasComputeType_t``, and ``hipblasOperation_t`` enumerations,
-see the `hipblas-common <https://github.com/ROCm/hipBLAS-common/blob/develop/library/include/hipblas-common/hipblas-common.h>`_ GitHub repository.
+see ``hipblas-common.h`` in the `hipblas-common GitHub <https://github.com/ROCm/hipBLAS-common>`_ repository.
 
 Definitions
 ===========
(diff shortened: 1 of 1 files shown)


TorreZuk:
Well I would wait on this and cherry pick thoughts until we know more

docs/reference/hipblas-api-functions.rst
```diff
@@ -147,7 +147,7 @@ hipBLAS types
 *************
 
 For information about the ``hipblasStatus_t``, ``hipblasComputeType_t``, and ``hipblasOperation_t`` enumerations,
-see the `hipblas-common <https://github.com/ROCm/hipBLAS-common/blob/develop/library/include/hipblas-common/hipblas-common.h>`_ GitHub repository.
+see ``hipblas-common.h`` in the `hipblas-common GitHub <https://github.com/ROCm/hipBLAS-common>`_ repository.
```
amcamd, line 151: Should this 
```see ``hipblas-common.h`` in the `hipblas-common GitHub` ```
change to 
```see ``hipblas-common.h`` in the `hipBLAS-common GitHub` ```

The difference is hipblas-common -> hipBLAS-common

amcamd (approved):
Possible capitalization change.

TorreZuk (approved):
While I don't see a good reason for the include tree change they don't want to revert so this looks like the best solution.

docs/reference/hipblas-api-functions.rst
amd-jnovotny: I'll fix this and push my change

stellaraccident:
Thanks guys - in a project like this, it is critically important that release docs don't point to develop branches. Good find, and let's fix the root cause.

(The header location was wrong. We're not changing it back. Develop moves forward)

If you'd like to discuss more, we've got office hours in an hour and can resolve this pretty quickly.

PR #999, opened 2025-04-30
remove mf16c and use C++11 _Float16

amcamd:
- rocBLAS uses the _Float16 datatype added in C++11
- hipblasHalf was using typedef to either 
  - __half from HIP
  - uint16_t
- it was using the compile flag -mf16c and the functions _cvtss_sh() and _cvtss_sh() to convert from float_to_half and half_to_float
- **This PR removes** 
  -**mf16c flag**
  -**use of _cvtss_sh() and _cvtss_sh()**
- One complication is that iostream cannot process _Float16, so the iostream operator << is overloaded in file clients/include/argument_model.hpp. The overload casts from _Float16 to float.

### DIFF
--- a/library/include/hipblas.h
+++ b/library/include/hipblas.h
@@ -100,18 +100,7 @@
 /*! \brief hipblasHanlde_t is a void pointer, to store the library context (either rocBLAS or cuBLAS)*/
 typedef void* hipblasHandle_t;
 
-/*! \brief To specify the datatype to be unsigned short */
-
-#if __cplusplus < 201103L || !defined(HIPBLAS_USE_HIP_HALF)
-
-typedef uint16_t hipblasHalf;
-
-#else
-
-#include <hip/hip_fp16.h>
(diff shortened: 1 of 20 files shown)


library/include/hipblas.h
```diff
@@ -100,18 +100,7 @@
...
-
-typedef uint16_t hipblasHalf;
-
-#else
-
-#include <hip/hip_fp16.h>
-typedef __half hipblasHalf;
-
-#endif
+typedef _Float16 hipblasHalf;
```
TorreZuk, line 103: It isn't really part of c++11 standard it is an extension from what I read.
I would keep the 
#if _cplusplus < 201103L 
as for API exposure with C/early C++ the uint16_t should suffice as before?

TorreZuk (approved):
Looks good if all tests pass

TorreZuk:
Looks like for g++ you need alternative.

PR #1000, opened 2025-05-01
Use permalink for hipblas-common.h link

davidd-amd:
When providing links to files in other projects provide a permalink rather than referencing a branch etc. There are no guarantees that files will never move. To get a permalink navigate to the file in github for the given branch/release/etc and then press "y" and the url will change to a permalink.


TorreZuk (commented):
Too much maintenance and so would become stale

amd-jnovotny:
We're going to link to the GitHub repository (root). Closing this as no longer required.

PR #1008, opened 2025-05-06
Setup Fortran on Windows.

lucbruni-amd:
Summary of proposed changes:
- Remove check against Windows during Fortran setup

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,12 +42,11 @@ if( NOT DEFINED CMAKE_CONFIGURATION_TYPES AND NOT DEFINED CMAKE_BUILD_TYPE )
   set( CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel." )
 endif()
 
-if (NOT WIN32)
-    if ( NOT DEFINED CMAKE_Fortran_COMPILER AND NOT DEFINED ENV{FC} )
-      set( CMAKE_Fortran_COMPILER  "gfortran" )
-    endif()
-    set( fortran_language "Fortran" )
-endif( )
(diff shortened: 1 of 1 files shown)


daineAMD (commented):
I'm fine with this, but is there a plan to enable fortran tests on windows as well? Would be good for everything to go in at once so the API doesn't get enabled for the testing.

lucbruni-amd:
> I'm fine with this, but is there a plan to enable fortran tests on windows as well? Would be good for everything to go in at once so the API doesn't get enabled for the testing.

@ScottTodd do you have an answer for this?

ScottTodd:
Ah. I had some notes on https://github.com/ROCm/TheRock/pull/515 that didn't make their way to this patch:

> Needed a few patches:
> * Find Fortran, [as in rocSPARSE](https://github.com/ROCm/rocSPARSE/blob/df47255d357badea8dac8d242a773c74eb89b593/clients/CMakeLists.txt#L46-L56) (hipBLAS may want a similar `BUILD_FORTRAN_CLIENTS` variable, or it could use the existing `BUILD_CLIENTS` variable (neither are _options_, they are variables that happen to get defined late in the file, or are set in toolchain/rmake files)

The change I made at the time was minimal to get hipBLAS building on Windows within TheRock, where we already set up Fortran and expect it to be available. I think we'll actually want something like what rocSPARSE does to keep Fortran client builds optional, but that will need some deeper plumbing than what this patch currently does.

To the question about Fortran testing, yes, I assume that we will enable Fortran tests on Windows too.

daineAMD:
Had to make some fixes to hipblas-common to get the math-ci windows pipeline running again, just kicked off a re-run of this PR and once CI all passes I'll approve to unblock TheRock. There's likely a decent amount of work to get testing enabled as well

daineAMD (approved):

ammallya:

PR #1010, opened 2025-05-07
Install `libhipblas_fortran.so`

lucbruni-amd:
This is required by the test and benchmark clients but was not part of the installation so far.


marbre:
This might not be needed, depending on where #989 goes.

lucbruni-amd:
> This might not be needed, depending on where #989 goes.

Converting to draft PR until 989 is concluded.

ammallya:

PR #1015, opened 2025-05-09
explicit install rpath with ORIGIN

TorreZuk:
* avoid packaging errors for invalid RUNPATH
* just a test for now as properties rely on llvm openmp


TorreZuk:
Changes merged in separate PR #1038

PR #1016, opened 2025-05-12
ci: remove junit step from test stage

samjwu:
attempt to resolve Math CI job running indefinitely

Summary of proposed changes:
- remove junit step from test stage for Math CI


TorreZuk (approved):
Did the junit step do nothing of value?

samjwu:
junit added graph of test results, but we suspect it's causing hangs

image

PR #1017, opened 2025-05-13
add back pre version 3.0 HIPBLAS_[RC] macros

amcamd:
This adds back macros used before hipBLAS version 3.0. It is best to use the new macros, but this will allow for use of the old macros for backward compatibility

### DIFF
--- a/library/include/hipblas.h
+++ b/library/include/hipblas.h
@@ -77,25 +77,6 @@
 #endif
 #endif
 
-// Macros used before hipBLAS version 3.0
-#define HIPBLAS_R_16F HIP_R_16F
-#define HIPBLAS_R_32F HIP_R_32F
-#define HIPBLAS_R_64F HIP_R_64F
-#define HIPBLAS_C_16F HIP_C_16F
-#define HIPBLAS_C_32F HIP_C_32F
-#define HIPBLAS_C_64F HIP_C_64F
-#define HIPBLAS_R_8I HIP_R_8I
-#define HIPBLAS_R_8U HIP_R_8U
-#define HIPBLAS_R_32I HIP_R_32I
-#define HIPBLAS_R_32U HIP_R_32U
-#define HIPBLAS_C_8I HIP_C_8I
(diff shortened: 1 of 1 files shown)


library/include/hipblas.h
```diff
@@ -77,6 +77,25 @@
 #endif
 #endif
 
+// Macros used before hipBLAS version 3.0
+#define HIPBLAS_R_16F HIP_R_16F
```
TorreZuk: I would put it at end of file to ensure they aren't used again in this file.   
@daineAMD  should they be in an #ifndef __HIP_PLATFORM_NVCC__ as well

daineAMD: Yeah good idea to move to the end. Seems like for this specific fix it could be excluded in cuda builds, not sure if something else will pop up, though.

TorreZuk: Or does HIP_R_16F get set up in cuda the same so fine for both?

daineAMD: Oh, yeah, this shouldn't give any issues with cuda, those types exist there too. So we can leave it for simplicity.

amcamd: next commit moves definitions to the end of hipblas.h

TorreZuk (approved):
Looks fine to help with those who ignored the V2 deprecations.

cgmb:
Thank you for keeping these a while longer. It can be difficult to maintain source compatibility with ROCm 5 without these compatibility shims. I'd probably end up duplicating these macros in application code otherwise.

> Looks fine to help with those who ignored the V2 deprecations.

Even if the deprecations are known, it can be difficult to maintain source compatibility with multiple versions of the library when breaking changes are made. You don't always have a place to put in an `#ifdef` if the sources are being hipified at build-time from CUDA code.

At some point, it will be reasonable to declare ROCm 6 a minimum requirement, but Ubuntu 24.04 is widely used and it ships with libhipblas-dev (5.5.1-1). It's nice to be able to keep applications source-compatible with that for a little while longer, and I was finding it tricky without these.

PR #1007, opened 2025-05-06
Check if ROCBLAS_PATH var is set before using it.

lucbruni-amd:
resolves https://github.com/ROCm/TheRock/issues/530

Summary of proposed changes:
- Check if ROCBLAS_PATH var is set before using it for gtest, otherwise report an error

PR #1018, opened 2025-05-14
revert breaking commit

TorreZuk:
* failed CI testing, revert while revising

### DIFF
--- a/clients/gtest/CMakeLists.txt
+++ b/clients/gtest/CMakeLists.txt
@@ -192,11 +192,7 @@ if (WIN32)
     add_custom_command( TARGET hipblas-test POST_BUILD COMMAND ${CMAKE_COMMAND} ARGS -E copy ${file_i} ${PROJECT_BINARY_DIR}/staging/ )
   endforeach( file_i )
 
-  if(DEFINED ${ROCBLAS_PATH})
-    add_custom_command( TARGET hipblas-test POST_BUILD COMMAND ${CMAKE_COMMAND} ARGS -E copy_directory ${ROCBLAS_PATH}/bin/rocblas/library/ ${PROJECT_BINARY_DIR}/staging/library/)
-  else()
-    message(WARNING "ROCBLAS_PATH not set, tests may be missing .dlls")
-  endif()
+  add_custom_command( TARGET hipblas-test POST_BUILD COMMAND ${CMAKE_COMMAND} ARGS -E copy_directory ${ROCBLAS_PATH}/bin/rocblas/library/ ${PROJECT_BINARY_DIR}/staging/library/)
 endif()
 
 set_target_properties( hipblas-test PROPERTIES


amcamd (approved):

ScottTodd:
Please notify the authors of changes when reverting their commits (I was co-author on https://github.com/ROCm/hipBLAS/pull/1007). This broke Windows builds in https://github.com/ROCm/TheRock. If we knew this revert was coming, we could have prepared a local patch or delayed updating hipBLAS instead of taking the breakage.

We'll need to work forward to a better fix that satisfies both projects. As it is now, this is broken code that tries to copy from files that do not exist like `/bin/rocblas/library` and files that should not be touched (`C:/Windows/System32/libomp140*.dll`).

ScottTodd:
Proposed a re-apply of the patch downstream: https://github.com/ROCm/TheRock/pull/656. When I have more time I can dig back into a general fix that works upstream and downstream.

TorreZuk:
> Please notify the authors of changes when reverting their commits (I was co-author on #1007). This broke Windows builds in https://github.com/ROCm/TheRock. If we knew this revert was coming, we could have prepared a local patch or delayed updating hipBLAS instead of taking the breakage.
> 
> We'll need to work forward to a better fix that satisfies both projects. As it is now, this is broken code that tries to copy from files that do not exist like `/bin/rocblas/library` and files that should not be touched (`C:/Windows/System32/libomp140*.dll`).

Sorry @ScottTodd I just mentioned the PR in the revert, I guess you are not watching the repo.   I will mention a co author or the whole list of co authors should be mentioned.  The original commit should never have been merged as broke the build in our CI.   As we are nearing branch creation I reverted as our policy.   The original intent I think I have restored in https://github.com/ROCm/hipBLAS/pull/1020 if it was just incorrect cmake syntax.   Is the azure pipeline for your build published so I can confirm?

ScottTodd:
> Sorry @ScottTodd I just mentioned the PR in the revert, I guess you are not watching the repo. I will mention a co author or the whole list of co authors should be mentioned.

Thanks! I wish GitHub had more built in alerts for such things.

> The original commit should never have been merged as broke the build in our CI. As we are nearing branch creation I reverted as our policy. The original intent I think I have restored in #1020 if it was just incorrect cmake syntax.

Ahhh that CMake syntax 🤦 . Thanks. 

> Is the azure pipeline for your build published so I can confirm?

I can test locally. We're still catching the CI/CD up to build all projects, which is why we missed this.
* Test workflow: https://github.com/ROCm/TheRock/blob/main/.github/workflows/build_windows_packages.yml (note `-DTHEROCK_ENABLE_BLAS=OFF` still, https://github.com/ROCm/TheRock/issues/543 to enable all projects)
* We're working on adding a trigger to test any ROCm subproject commit: https://github.com/ROCm/TheRock/issues/559

PR #1020, opened 2025-05-14
ROCBLAS_PATH optional

TorreZuk:
* restore reverted PR
* toolchain file always sets it so this is custom build
* insitu testing will still fail so this may not be desirable


ScottTodd (approved):
Thanks! Tested locally in TheRock, where we don't set `ROCBLAS_PATH`. Our build continues to work as expected with this change.

rkamd (approved):

PR #1027, opened 2025-06-04
define macro HIPBLAS_V2, it was used in hipblasVersionMajor==2 for de…

amcamd:
- define HIPBLAS_V2 in case other projects are using it in a condition like:
```
#if defined(USE_ROCM) && ((ROCM_VERSION_MAJOR == 6 && defined(HIPBLAS_V2)) || ROCM_VERSION_MAJOR >= 7)
```

### DIFF
--- a/library/include/hipblas.h
+++ b/library/include/hipblas.h
@@ -25861,6 +25861,7 @@ HIPBLAS_EXPORT const char* hipblasStatusToString(hipblasStatus_t status);
 #endif
 
 // Macros used before hipBLAS version 3.0
+#define HIPBLAS_V2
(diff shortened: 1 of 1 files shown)


TorreZuk (approved):
Note the compiler define HIPBLAS_V2 was under their control, it was users choice to set it so don't see strong case for us to define it for everyone.  However as it is prefixed with our namespace think it is acceptable.

TorreZuk:
Also if they did the right thing and set the define I think a redefinition error will be triggered if they don't remove.  But I guess we could argue this helps them clean out their configuration?

daineAMD (approved):

PR #1028, opened 2025-06-06
cherry-pick addition of HIPBLAS_V2

amcamd:
- cherry pick addition of macro HIPBLAS_V2 to hipblas.h
- This could have been used in cpp logic like
```
#if defined(USE_ROCM) && ROCM_VERSION >= 60000 && defined(HIPBLAS_V2)
  //use hipBLAS version 3 API
#else
  // use hipBLAS API before version 3
# endif
```
It would be more correct to use
```
#if (defined HIPBLAS_V2 && hipblasVersionMajor == 2) || hipblasVersionMajor >= 3
  //use hipBLAS version 3 API
#else
  // use hipBLAS API before version 3
# endif
```

### DIFF
--- a/library/include/hipblas.h
+++ b/library/include/hipblas.h
@@ -25861,6 +25861,9 @@ HIPBLAS_EXPORT const char* hipblasStatusToString(hipblasStatus_t status);
 #endif
 
 // Macros used before hipBLAS version 3.0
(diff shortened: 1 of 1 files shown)


TorreZuk (approved):
LGTM

daineAMD (approved):

PR #1032, opened 2025-06-16
SWDEV-531400 Remove File reorganization backward computability

jonatluu:
Summary of proposed changes:
remove backwards compatibility code. Feature has been disabled related code is not needed.
-
-
-


TorreZuk (approved):
Thanks

PR #1036, opened 2025-06-19
include stddef for offsetof

amcamd:
- the c function offsetof is defined in stddef.h
- this function offsetof is used in library/include/hipblas.h
- include stddef.h in hipblas.h to provide the definition for offsetof


TorreZuk (approved):
Clang format I would think needs it up one item in the list

amcamd:
duplicate of https://github.com/ROCm/hipBLAS/pull/1035

PR #1038, opened 2025-06-20
Openmp cmake config

estewart08:
Summary of proposed changes:
- First search for ROCm's libomp.so via openmp-config.cmake. This is what we would prefer instead of searching for a system libomp.so/libgomp.so and then manually adding in a ROCm lib path.
- This methodology should still be RHEL-10 RPATH compliant.
- This patch is applied ontop of #1015

### DIFF
--- a/clients/CMakeLists.txt
+++ b/clients/CMakeLists.txt
@@ -58,6 +58,18 @@ function( get_cblas cblas_libs cblas_inc )
   set( ${cblas_libs} ${libs} PARENT_SCOPE )
 endfunction( )
 
+function( apply_omp_settings lib_target_ )
+  if (TARGET OpenMP::OpenMP_CXX)
+    set_target_properties( ${lib_target_} PROPERTIES
+      BUILD_RPATH "${HIP_CLANG_ROOT}/lib"
+    )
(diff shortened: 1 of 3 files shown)


TorreZuk (approved):
It looks good but need to test manually a bit to confirm.  Can teams message for docker or other build to test against.

estewart08:
This also will take care of a future situation where `LLVM_ENABLE_PER_TARGET_RUNTIME_DIR=ON` is used when building llvm-project. When that option is turned on, libomp.so is installed into `/opt/rocm/lib/llvm/lib/x86_64-unknown-linux-gnu`. The openmp-config.cmake will be automatically updated with the new path.

TorreZuk:
@estewart08  this seems to work in my latest docker so are we still good to merge and cherry pick this?

estewart08:
@TorreZuk, that is fine with me. Not sure if the failing checks need to run again.

TorreZuk:
@estewart08 not sure on the Azure failures, can you look as maybe the old cmake config is bad so causes the errors shown in the build log?    Not sure how our other builds are all fine.  Maybe rebase PR on develop and run again.

PR #1049, opened 2025-07-17
gcc default OpenMP

TorreZuk:
* use old system for gcc build

### DIFF
--- a/clients/CMakeLists.txt
+++ b/clients/CMakeLists.txt
@@ -59,7 +59,7 @@ function( get_cblas cblas_libs cblas_inc )
 endfunction( )
 
 function( apply_omp_settings lib_target_ )
-  if( CMAKE_CXX_COMPILER_ID MATCHES "Clang"  )
+  if( CMAKE_CXX_COMPILER_ID MATCHES "Clang" )
     if (TARGET OpenMP::OpenMP_CXX)
       set_target_properties( ${lib_target_} PROPERTIES
         BUILD_RPATH "${HIP_CLANG_ROOT}/lib"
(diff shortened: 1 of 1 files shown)


daineAMD (approved):

clients/CMakeLists.txt
```diff
@@ -134,9 +136,11 @@ if( BUILD_CLIENTS_BENCHMARKS OR BUILD_CLIENTS_TESTS)
   find_package(Threads REQUIRED)
 
   # Look for openmp config in ROCm install to populate openmp_LIB_DIR and openmp_LIB_INSTALL_DIR
-  find_package(OpenMP CONFIG PATHS "${HIP_CLANG_ROOT}/lib/cmake")
+  if( CMAKE_CXX_COMPILER_ID MATCHES "Clang"  )
+    find_package(OpenMP CONFIG PATHS "${HIP_CLANG_ROOT}/lib/cmake")
+  endif()
 
-  if (TARGET OpenMP::omp)
+  if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND TARGET OpenMP::omp)
```
estewart08, line 141: This "clang check" here may be a bit repetitive as the only way OpenMP::omp is set if line 140 is executed.

clients/CMakeLists.txt
```diff
@@ -59,20 +59,22 @@ function( get_cblas cblas_libs cblas_inc )
...
-      INSTALL_RPATH "$ORIGIN/../llvm/lib"
-    )
-  elseif(TARGET OpenMP::omp)
-    set_target_properties( ${lib_target_} PROPERTIES
-      BUILD_RPATH "${HIP_CLANG_ROOT}/${openmp_LIB_DIR}"
-    )
-    set_target_properties( ${lib_target_} PROPERTIES
-      INSTALL_RPATH "$ORIGIN/../llvm/${openmp_LIB_DIR}"
-    )
+  if( CMAKE_CXX_COMPILER_ID MATCHES "Clang"  )
```
estewart08: We cannot guard `TARGET OpenMP::OpenMP_CXX` with `Clang` because below we are using -L /path/to/rocm/lib -lomp which ldd shows libomp.so -> not found. We would need to move the Clang guard down to line 70.

`elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND TARGET OpenMP::omp)`

clients/CMakeLists.txt
TorreZuk: > This "clang check" here may be a bit repetitive as the only way OpenMP::omp is set if line 140 is executed.

Sure but like to be clear when I expect to going into the if

clients/CMakeLists.txt
TorreZuk: See if later revisions covers this now, the rpaths are all llvm/clang properties so think fine to do nothing for gcc libgomp.

PR #1047, opened 2025-07-14
CP merge pull request #1038 from estewart08/openmp-cmake-config (#1043)

TorreZuk:
OpenMP cmake config support along with RPATH updates.

(cherry picked from commit fe3af3c53d41394904d893dcbec4d00165220955)

(cherry picked from commit ad2e8cb9ee42883238afd5e7a494f7a2774b3807)

PR #1051, opened 2025-07-24
Revert "CP merge pull request #1038 from estewart08/openmp-cmake-config (#1043)"

amcamd:
Reverts ROCm/hipBLAS#1047 to unblock, CI build failure, https://ontrack-internal.amd.com/browse/SWDEV-546120

### DIFF
--- a/clients/CMakeLists.txt
+++ b/clients/CMakeLists.txt
@@ -58,25 +58,6 @@ function( get_cblas cblas_libs cblas_inc )
   set( ${cblas_libs} ${libs} PARENT_SCOPE )
 endfunction( )
 
-function( apply_omp_settings lib_target_ )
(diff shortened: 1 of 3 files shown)


pramenku:
I see math CI is good, http://math-ci.amd.com/job/main/job/precheckin/job/hipBLAS/job/PR-1051/1/pipeline-overview/?selected-node=773
 
UB22 one is failed which is harbor cert issue on the build machine. Other UB22, rhel9 and sles is GREEN

TorreZuk:
The failure looked to be related to an outdated dependency, this revert may break other things.  Correct fix would be to use AOCL gcc 4.2 to resolve.

pramenku:
> The failure looked to be related to an outdated dependency, this revert may break other things. Correct fix would be to use AOCL gcc 4.2 to resolve.

@TorreZuk Let's revert this then. Make sure staging is done on it as it was never done before.

Summary: Hipblas Change requires AOCL4.2 which was missing in UB24 hence failing

PR #1058, opened 2025-09-24
Disable dependabot.yml

samjwu:
Disables dependabot PRs

Summary of proposed changes:

- stops dependabot PRs for docs
- dependabot.yml is separate from the settings/advanced security config, but still has to be valid.
- similar to https://github.com/ROCm/rocBLAS/pull/1670


TorreZuk (approved):
If working in rocBLAS should work here

